PR #1, opened 2021-06-18
MAINT: add example

andyfaff:
- add example
- flake8 the code
- add .gitattributes to normalise line endings
- update AUTHORS.rst


arm61:
Is there a reason for those linting ignores?

andyfaff:
I just blindly copied them over from what I used for refnx. I'll remove them now. I think I used them back before I used black.
(I like using black, but it can be a bit dictatorial).

andyfaff:
@aglavic do you have the ORT file you used during the tutorial, can you add it to the repo?

PR #2, opened 2021-06-22
adding a draft of the base objects

arm61:
A first draft of the base classes to define the ORSO files. 

In particular I am curious about the thoughts on: 
1. The use of `_check_unit`, is it maybe more sensible to have a superclass for all things with units?
2. How do we want to describe a vector direction? I have this as a string currently, is this relevant?


arm61:
Anyone is welcome to review btw. Not expecting everyone to review either.

aglavic:
To your questions:
1. I'll go through the code and maybe comment on it directly.
2: Yes, vectors are relevant e.g. for polarization analysis and external field direction. I suggest x||beam@0°, z||surface normal. We should use a list of three numbers instead of a string to easily convert to array.

orsopy/fileio/base.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def noop(self, *args, **kw):
+    pass
+
+
+yaml.emitter.Emitter.process_tag = noop
+
+
+def _repr(class_to_represent):
```
aglavic, line 17: Why do you define this as a function and not a method of a base class that is inherited?

arm61:
Just trying to understand your notation in 2. is that "x parallel to beam at 0 degree" and "z parallel to surface normal"? what is the third number you mention in the next sentence?

I think it makes the most sense for a `MagnetizationVector` class (like this https://github.com/reflectivity/orsopy/blob/cf73b31ef9555b2aa465ff79dd238e8d4d699525/orsopy/fileio/base.py#L47) that can be used in the `ValueVector` classes. Keeping the latter general.

orsopy/fileio/base.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+
+    def __repr__(self):
+        """
+        The string representation for the Header class objects.
+        """
+        return _repr(self)
+
+
+class ValueScalar(Header):
```
aglavic: I'm not sure why we need such a large class structure. In principle a dictionary or derived object would be sufficient that then just checks against the schema file. In this way we only have to update a scheme and not the code here.
For me this would be quite a lot of additional overhead when using the package to put all my values first in one of these classes upon writing and extract the values from the classes when reading. It seems simpler to just supply a full dictionary and get an error message if it does not follow the schema. (Warning for recommendations, error for missing data or wrong entries.)

In case you like to stick to custom classes for everything, any kind of checks should be methods of the associated classes. Then the classes also need a facility to check for validity in certain entries against the schema. (e.g. if values and value ranges are allowed).

Also keep in mind that the library should be used by other programs, so it would be good to make sure resulting output is available in standard types. (str, float, int, list, dict and ndarray)

acaruana2009:
1. I think both solutions work, although I feel like having the method `_check_unit()` as a class method inside say a unit class might be a cleaner way to go in the long term. 
2. I agree with Artur about the vectors. We just need to agree a convention. The convention Artur has suggested is fine with me.
x= along the beam, y=across the beam but perp to z

How do we want to define the polarisation? `+` = guide field || pol vector, and `-` = guide field **anti** || pol vector?

orsopy/fileio/base.py
```diff
@@ -0,0 +1,174 @@
...
+        class_to_represent (:py:class:`object`): The class to be represented.
+
+    Returns:
+        (:py:attr:`str`): A string representation.
+    """
+    cleaned_data = dict(
+        (k, v) for (k, v) in class_to_represent.__dict__.items(
+        ) if v is not None
+    )
+    return yaml.dump(cleaned_data, sort_keys=False)
```
aglavic, line 32: Shouldn't the representation be Pyhon-like. So something like "Class(attr1=value, attr2=value2)".

For the dump to yaml I would use a "dump" method (with possible keyword parameters).

orsopy/fileio/base.py
```diff
@@ -0,0 +1,174 @@
...
+    Args:
+        magnitude (:py:attr:`float`): The value.
+        unit (:py:attr:`str`, optional): The unit. Optional,
+            defaults to :code:`'dimensionless'`.
+    """
+
+    def __init__(self, magnitude, unit="dimensionless"):
+        _check_unit(unit)
+        self.magnitude = magnitude
+        self.unit = unit
```
aglavic: I think units should be checked against the allowed ones. (In the class for generally being allowed and in later structures for the schema.)

orsopy/fileio/__init__.py
```diff
@@ -0,0 +1,3 @@
+from orsopy.fileio.base import *
```
aglavic, line 1: I would not use a * import in a package file.

arm61:
I have added a `Unit` class, cause I agree on the clean point. For the magnetisation vector, that's two parameters then?

tests/test_fileio/test_base.py
```diff
@@ -0,0 +1,103 @@
+"""
+Tests for fileio.base module
+"""
+
+# author: Andrew R. McCluskey (arm61)
+
+from orsopy.fileio.base import ValueVector
+import unittest
+from numpy.testing import assert_almost_equal, assert_equal
```
aglavic, line 9: For non-array objects I would just use the assert attributes of the TestCase class. We should definitely  check that floats/ints are actually floats and not numpy scalar objects, as these lead to issues with yaml when writing.

aglavic (commented):
I hope this helps, let me know if you want to discuss any of it.

aglavic:
> Just trying to understand your notation in 2. is that "x parallel to beam at 0 degree" and "z parallel to surface normal"? what is the third number you mention in the next sentence?
> 
> I think it makes the most sense for a `MagnetizationVector` class (like this
> 
> https://github.com/reflectivity/orsopy/blob/cf73b31ef9555b2aa465ff79dd238e8d4d699525/orsopy/fileio/base.py#L47
> 
> ) that can be used in the `ValueVector` classes. Keeping the latter general.

I don't think we need a magnetization vector, just a standard 3D vector should do. The third one is y and given by the other two.

aglavic:
> How do we want to define the polarisation? `+` = guide field || pol vector, and `-` = guide field **anti** || pol vector?

Suppose this is the standard definition, so we can follow it.

bmaranville:
Instead of using comment type hints, can we use the newer style of type hints?  Then libraries are available to do data validation of the kind you are building in as functions, e.g. these are the classes I wrote to generate the schema earlier:

import datetime
import enum
from typing import Optional, Union, List, Literal, Dict, Any
from dataclasses import field, dataclass

def d(t):
    return field(metadata={"description": t})

@dataclass
class Creator:
    name: str
    affiliation: str
    time: datetime.datetime = d("timestamp string, formatted as ISO 8601 datetime")
    system: str

@dataclass
class Sample:
    name: str

@dataclass
class Experiment:
    instrument: str
    probe: Union[Literal['neutron', 'xray']]
    sample: Sample

class Polarisation(str, enum.Enum):
    """ The first symbol indicates the magnetisation direction of the incident beam.
    An optional second symbol indicates the direction of the scattered beam, if a spin analyser is present."""
    p = '+'
    m = '-'
    mm = '--'
    mp = '-+'
    pm = '+-'
    pp = '++'

@dataclass
class Value:
    magnitude: Union[float, List[float]]
    unit: Optional[str] = field(default=None, metadata={"description": "SI unit string"})

@dataclass
class ValueRange:
    min: float
    max: float
    steps: Optional[int] = None
    unit: Optional[str] = field(default=None, metadata={"description": "SI unit string"})

@dataclass
class Measurement:
    scheme: str
    omega: Union[Value, ValueRange] = d("probe angle of incidence")
    wavelength: Union[Value, ValueRange]
    polarisation: Optional[Polarisation] = None

@dataclass
class DataSource:
    owner: str
    facility: str
    experimentID: str
    experimentDate: str
    title: str
    experiment: Experiment
    measurement: Measurement


@dataclass
class ORSOHeader:
    creator: Creator
    data_source: DataSource

________________________________

tests/test_fileio/test_base.py
```diff
@@ -0,0 +1,103 @@
...
+from orsopy import fileio
+
+
+class TestHeader(unittest.TestCase):
+    def test_value_scalar(self):
+        c = fileio.ValueScalar(1., 'm')
+        assert_almost_equal(c.magnitude, 1)
```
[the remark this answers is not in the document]
arm61: Do we want to check type?

orsopy/fileio/base.py
arm61: My motivation for the class structure was two fold: 
- making checking easier
- making documentation clearer/easier

I am very open to changing my mind, but this was how I designed the amor code so I was going off of that. 

There is no reason we can't have a helper function(s) that consumes/produces a dictionary for the full header.

arm61:
> > Just trying to understand your notation in 2. is that "x parallel to beam at 0 degree" and "z parallel to surface normal"? what is the third number you mention in the next sentence?
> > I think it makes the most sense for a `MagnetizationVector` class (like this
> > https://github.com/reflectivity/orsopy/blob/cf73b31ef9555b2aa465ff79dd238e8d4d699525/orsopy/fileio/base.py#L47
> > 
> > ) that can be used in the `ValueVector` classes. Keeping the latter general.
> 
> I don't think we need a magnetization vector, just a standard 3D vector should do. The third one is y and given by the other two.



> > Just trying to understand your notation in 2. is that "x parallel to beam at 0 degree" and "z parallel to surface normal"? what is the third number you mention in the next sentence?
> > I think it makes the most sense for a `MagnetizationVector` class (like this
> > https://github.com/reflectivity/orsopy/blob/cf73b31ef9555b2aa465ff79dd238e8d4d699525/orsopy/fileio/base.py#L47
> > 
> > ) that can be used in the `ValueVector` classes. Keeping the latter general.
> 
> I don't think we need a magnetization vector, just a standard 3D vector should do. The third one is y and given by the other two.

My motivation for the more specific class object was to document the specification of which is x/y/z.

bmaranville:
I agree that agreeing on the definition of reference for the magnetisation vector is important and that it should get its own class in this context (that was an awkward sentence).
________________________________

orsopy/fileio/base.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    Checks if the input is ASCII.
+
+    Args:
+        s (:py:attr:`str`): Input to check
+
+    Returns:
+        (:py:attr:`bool`): True if value is ASCII
+    """
+    return all(ord(c) < 128 for c in s)
```
andyfaff, line 173: [`str.isascii`](https://docs.python.org/3/library/stdtypes.html#str.isascii)

andyfaff:
It's good to see work progressing on this. However, before we go much further I'm wondering if it'd be wise for a sketch to be made of the proposed architecture. Either in pencil (or a drawing package), or as a series of empty classes that will show the API.

e.g. 
```
class ORSO:
      # this class is going to do x.
      
     def functionality(self):
          # this method is going to do this specific task
```

At the moment I don't know what we all have in mind for the design, or how what's contained in this PR relates to the eventual goal. I guess what I'm asking for is for us to figure out first how the package will be used, which will guide the API, which will then guide the nuts and bolts, this PR seems to be lower level nuts and bolts.

arm61:
My aim with this code was to have all of the base objects that we need to then build the more complex structure of the ORSO header. I have an older version of the header builder here (https://scipp.github.io/ess/techniques/reflectometry/orso.html), which I was taking as inspiration. But I was trying to do things gradually (to better understand others' expections).

arm61:
Just for added information. The way I wrote the older version, the string representation for the `ORSO` class object (which contains all of the other objects) output the header. So when writing the file it was as simple as 
```
np.savetxt(filename, np.array([q_z_vector, intensity, dintensity, dq_z_vector]).T, fmt='%.16e', header=str(header))
```

aglavic:
> It's good to see work progressing on this. However, before we go much further I'm wondering if it'd be wise for a sketch to be made of the proposed architecture. Either in pencil (or a drawing package), or as a series of empty classes that will show the API.
> 

I agree, that we seem to have different concepts of the end product in mind. Maybe it would be helpful to have a face to face discussion to settle on a general architecture?

From my point of view the library should have the following priorities that the architecture should fulfill:
1. Correctness of result, stay within ORSO specification
2. Pure python, no unnecessary clutter. (e.g. just yaml and numpy from non standard library modules)
3. Ease of use to read data, single class/function that returns easy to use objects
4. Ease of use to write data, best case would be usability without any look into our API reference
5. Clear code to be maintainable
6. Performance of reader (and maybe writer)
7. Keep usage pythonic, e.g. access attributes like in dictionaries etc.

arm61:
I think that a "f2f" meeting is a good idea. How do we wanna arrange that?

aglavic:
I guess we can have a quick poll.
https://doodle.com/poll/mhhqkhfk7rwc9qkt?utm_source=poll&utm_medium=link

Can you send this to everyone who should be invited and we decide until the weekend for the appropriate slot?

arm61:
Looks like just @bmaranville and @andyfaff are the only ones that would be interested that haven't done the poll (although I doubt the times will be able to catch both of them).

bmaranville:
I put a dataclass-based version of base.py in the dataclass_base branch, which I just pushed.  There is a test in tests/test_header.py that will generate a header from a class structure using those dataclasses.

The schema used is not up to date - it is based on the first example file sent around that the original schema json files were generated from, but it provides an example of coding with the dataclasses and ways to generate dicts from it.

bmaranville:
It turns out that a third-party library (pip install dataclasses-json) allows very easy deserialization, where you can just read in the YAML header and call ORSOHeader.from_dict(header) to re-read your header into ORSO classes.

An example is posted in the dataclass_base branch in tests/test_deserialize (it requires that the schema classes also be decorated with @dataclass_json, though this can be mapped to a no-op if the library is not found, and it will have no effect on the serialization)

arm61:
Nice find. I will try and bring this all together tomorrow/next week to make something. Btw, I just clicked in the reason for your T-shirt choice today 🇨🇦

arm61:
I have done everything that we outlined yesterday (I think, my notes might have been off). I didn't add the schema generation parts as this isn't necessary until we have something more than these base classes. 

One of the results of my automatic identification of optional arguments is that we are limited to py3.8 and py3.9, I am okay with that, but wanna see if someone has a better way to do this (the automatic identification of optionals)?

bmaranville:
I think your test for Optional is good, but for attributes that are optional, do we want to define default values?  That would mean that in order to have a value "cleaned" you would have to specify it as None, rather than not specifying it.  This seems somewhat non-standard behavior.

Also, for internal bookkeeping, you could add an "write_if_None": False item to the field metadata.  It seems somewhat redundant but in this case I think it's really not - Optional[int] really means nullable as a type, and our "write_if_None" flag means both "it can be None and don't write it to the dict when it is None".  There are cases in our schema where the type is a Union with None, but it is not an optional field for writing to the header. 
I do realize that "write_if_None" is a terrible name for a key.

arm61:
You make a good point on the optional values thing, this has been changed in the latest commit. 

I am not sure I get your `write_if_None` thing. You want a flag in the metadata for Optional items that means the Optional item will be written even if it is None? In that case, shouldn't the value just be non-Optional then (it becomes a question of how flexible vs prescriptive we wanna be I think). 

On that last thing, I can't seem to get the value of the attribute back when using `fields()`, I might be missing something though (very new to these python data class types).

bmaranville:
The write_if_None flag is to address the ambiguity in what "Optional" means... from the python typing perspective, that just means it can have a value of None.  From the ORSO perspective, we want to be able to specify some fields that are "required" but could potentially have a value of None, as far as I recall, so using Optional typing for both won't work.

The dataclasses.fields(\<dataclass\>) method only gives information about the field definitions, not the values stored in them.

arm61:
The way I had it in my mind was, say for person you could define as
```
p1 = Person('A Person', 'A Uni')
```
Which produces the`p1.to_yaml()`
```
name: A Person
affiliation: A Uni
```
Since the email is (ORSO) optional we don't bother to print it if it isn't given. If it is given it would be,
```
p2 = Person('Another Person', 'A Different Uni', email='another.person@adiffuni.edu')
```
Then `p2.to_yaml()` gives:
```
name: Another Person
affiliation: A Different Uni
email: another.person@adiffuni.edu
``` 
But if something that was (ORSO) required wasn't available (say affiliation) we would have the following,
```
p3 = Person('An Recent Graduate', None)
```
Which gives the yaml
```
name: An Recent Graduate
affiliation: null
```
This logic probably is a bit duck-typed, cause the schema is being ignored which I suspect is your issue?

bmaranville:
You are correct - my issue is that if we are using typing.Optional to flag ORSO-optional values, then we can't also use it for validation of ORSO-required values that might have a value of None.  My suggestion is that we use the typing types for validation (since there are 3rd-party libraries that are already set up to do this) and flag the ORSO-optional values in some other way.

arm61:
What do you think of something like the `orso_optionals` attribute that I have added?

bmaranville:
That looks like a reasonable solution - maybe the most straightforward.  I had wondered about special decorators or the like but the list seems more obvious to readers of the code.

arm61:
For the third-party library interoperation for validation does the typing need to change now?

bmaranville:
The typing seems pretty good - this could be used for validation as-is, I think.

arm61:
I think this is ready to merge unless @andyfaff @acaruana2009 or @aglavic have any other comments? Then I will start on the next structures up.

PR #3, opened 2021-07-07
add the objects for the full range of classes

arm61:
Also:
- introduce recursion such that nested orsopy objects are correctly converted to yaml
- Add correct datetime formatting (for some reason pyyaml doesn't use the `T` delimitation


orsopy/fileio/__init__.py
```diff
@@ -1,4 +1,5 @@
 from orsopy.fileio.base import (ValueScalar, ValueRange, ValueVector, Comment,
                                 Person, Column)
+from orsopy.fileio.reduction import Software, Reduction
```
aglavic, line 4: I would always use the from .XXX import YYY form to make it possible/easy to include the package in other python projects directly.

orsopy/fileio/base.py
```diff
@@ -33,8 +44,14 @@ def _clean(self):
         :return: Cleaned dictionary
         :rtype: dict
         """
-        return dict((k, v) for (k, v) in self.__dict__.items()
-                    if (v is not None or k not in self.orso_optionals))
+        d = {}
+        for k, v in self.__dict__.items():
```
aglavic, line 48: This will only return any attributes set on the instance and ignore anything from the class and inherited from parents. I would expect that e.g. default arguments would therefore be ignored. I would propose:
`[(k, getattr(self, k)) for k in dir(self) if not k.startswith('_') and not callable(getattr(self, k))]`

This should include all non-hidden attributes of the instance, class and inherited and would even allow special cases of sub-classes that want to overwrite it by using the __dir__ method.

orsopy/fileio/base.py
```diff
@@ -33,8 +44,14 @@ def _clean(self):
...
+            if (v is not None or k not in self.orso_optionals):
```
aglavic, line 49: I think this could include callable objects if they are set during instantiation and are not in orso_optionals.

orsopy/fileio/reduction.py
```diff
@@ -0,0 +1,45 @@
...
+# author: Andrew R. McCluskey (arm61)
+
+from typing import Optional, List
+from dataclasses import field, dataclass
+import datetime
+from dataclasses_json import dataclass_json
+from .base import Header, Person
+
+
+@dataclass_json
```
aglavic, line 14: Do you use any of the functionality provided by this? E.g. it already provides the .to_dict method. If we do this our selves anyway, including this decorator might have undesired side-effects.

orsopy/fileio/base.py
```diff
@@ -33,8 +44,14 @@ def _clean(self):
...
+                if hasattr(v, 'orso_optionals'):
+                    d[k] = v._clean()
+                else:
+                    d[k] = v
+        return d
```
aglavic, line 54: btw. why is this a hidden method and called _clean? Isn't it the default dictionary representation?

I would use the name "to_dict" to make it similar to "to_json" or "to_yaml".

orsopy/fileio/reduction.py
arm61: I haven't tried reading files in, but my logic was that the `from_dict` method from `@dataclass_json` would be useful in file reading in the future.

orsopy/fileio/__init__.py
arm61: I have no opposition to this style, I just thought it would be clearer if the user could have a single import to get all of the data classes that are necessary.

orsopy/fileio/base.py
arm61: `to_dict` currently exists as a `@dataclass_json` method. I agree, if we don't use `@dataclass_json` but not if we are.

orsopy/fileio/reduction.py
arm61: @bmaranville had the suggestion of using `@dataclass_json` so perhaps he can elucidate more.

orsopy/fileio/base.py
arm61: A potential issue with using `dir` that it doesn't keep the same order as the attributes are define.

arm61: Turns out `self.__dir__()` and `dir(self)` return different things (https://stackoverflow.com/questions/9659361/order-of-results-from-pythons-dir-function/9659377)

orsopy/fileio/reduction.py
bmaranville: I included dataclass_json just as a proof-of-concept for reading headers back into the python classes - I agree that it probably doesn't belong in our standard library.

orsopy/fileio/base.py
bmaranville: dataclasses.asdict does this for you - it collects all the fields that have been defined, recursively through every attribute that is also decorated as a dataclass (so no inheritance is needed)

orsopy/fileio/reduction.py
arm61: Fair enough, I thought you were super keen on it but I guess I misread the room.

orsopy/fileio/base.py
arm61: It wasn't catching the `orso_optionals` from each inherited level for me, hence the recursive method.

orsopy/fileio/reduction.py
bmaranville: Oh, I do like it very much!  I just though we were trying to avoid third-party libraries in this code.

orsopy/fileio/base.py
arm61: Since we aren't using `@dataclass_json` I have changed `_clean` to `to_dict`.

arm61:
There was some issues with getting lists being parsed correctly, which has lead to the latests design of `to_dict()` (formerly known as `_clean()`. Thoughts on how to improve this are very welcome.

arm61:
Does anyone have strong feelings on Artur's comment here https://github.com/reflectivity/orsopy/pull/3#discussion_r665237231 ? If not I will follow is logic cause I am not fussed. 

Additionally I have put the `Orso` class object which bring everything together directly in the `__init__.py` in `fileio`, which people might disagree with.

orsopy/fileio/__init__.py
andyfaff: I normally have private classes and methods prefixed by _. I then add all public classes and functions to submodule.__all__. So the import would be from `orsopy.file import stuff`. For really important things you could add them to `orsopy.__all__`.

arm61: I am tempted towards Artur's approach in the first instance and we can evolve in future based on how we use it.

aglavic: > I have no opposition to this style, I just thought it would be clearer if the user could have a single import to get all of the data classes that are necessary.

I agree, that it is nice to have all in the package. My point was about source of the import. What I meant was to replace
`from orsopy.XXX` by `from .XXX` so that it works if orsopy is a sub-packge of another python program.

andyfaff: Ahh, that's a lot clearer

PR #4, opened 2021-08-09
Read the header

andyfaff:
AFAICT there is no header loader anywhere. This PR adds a loader for the header. Is this the right place to add it?

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -188,7 +188,7 @@ class File(Header):
             for line in fi.readlines():
                 if not line.startswith("#"):
                     break
-                header.append(line)
+                header.append(line.lstrip("#"))
             return ''.join(header)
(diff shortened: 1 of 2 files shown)


andyfaff:
`pattern = re.compile("^(# ORSO reflectivity data file \| ([0-9]+\.?[0-9]*|\.[0-9]+) standard \| YAML encoding \| https://www\.reflectometry\.org/)$")`

In place of my memory

arm61:
Yeah, I hadn't done anything on loading, was just putting together the base objects for the structure. That said, do we want this in the `File` class? My understanding was that class was really for describing all files (reduced, raw, etc.) so maybe we want super class of `File` for the reduced data files that includes this header loading functionality?

andyfaff:
Closed because of things added in #12

PR #15, opened 2021-08-12
ENH: bring closer to original definition

andyfaff:
@arm61, this PR is not the smallest. 

- The test_example.ort file now validates against the json schema
- A header produced by `orsopy.fileio.orso.ORSO` now validates against the json schema
- ORSO.measurement moved to ORSO.data_source.measurement to reflect the layout in the example file
- the json schema produced from @bmaranville script is being used. Eventually we should generate from the actual ORSO classes. But they're not super far apart at the moment.
- one can now import classes as `from orso.fileio import DataSource`.

I still have uncertainty about some aspects; specifically how times and dates are referred to. In this PR I've tried to remain faithful to https://www.reflectometry.org/projects/file_formats/tasks/ws_2021-06_text/. However, there is some variation. e.g. there is `ORSO.creator.time` and a `ORSO.data_source.measurement.data_files[].created`. Everything should probably be called one of `time/timestamp`.

My plan for tomorrow is to start playing around with the ability to construct an actual file with numerical data and the header, etc. As such I intend to merge this PR tomorrow.


arm61:
I thought we had gotten rid of linting for now?

andyfaff:
I can't remember the conversation on linting. By linting do you mean testing with flake8 on the code? (I can't remember the conversation on that). Or did you mean linting <-> validation?

I think having the ability to schema validate is important to make sure that the ort files have the correct structure, and the names of each of the nodes is the same as https://www.reflectometry.org/projects/file_formats/tasks/ws_2021-06_text/. So using a validator during testing is good, but users don't necessarily need to validate when reading something in. Using a validator also means that one doesn't have to check the output of `ORSO.to_yaml()` against a specific string, that's pretty fragile.

PR #16, opened 2021-08-13
ENH: file writer for an ORSO object

andyfaff:
This PR:

1. adds the `ORSO.save` method to write the object to file.
2. adds a data attribute to `DataSource`.
3. adds the `ORSO.add_data_source` method
4. tests that the ORSO header from the saved file is valid.
5. adds functionality to load the data-arrays from an ORSO file.
6. re-reads the ORSO file

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -220,7 +220,7 @@ def _read_header(file):
         header = []
         for line in fi.readlines():
             if not line.startswith("#"):
-                break
+                continue
             header.append(line[1:])
         return "".join(header)
 
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -14,6 +14,8 @@ import json
 import yaml
 from contextlib import contextmanager
 import re
+import numpy as np
+
 from .. import orsopy
 
 
@@ -213,16 +215,61 @@ class File(Header):
                 )
 
 
-def _read_header(file):
-    # reads the header of an ORSO file.
-    # does not parse it
+def _read_header_data(file):
+    """
+    reads the header and data contained within an ORSO file.
+    It does not parse the header
+
+    Parameters
+    ----------
+    file: str or file-like
+
+    Returns
+    -------
+    header, data_sets: str, list
+        `header` is a string containing the unparsed yaml header. This has to
+        be processed further
+        `data_sets` is a Python list containing numpy arrays holding the
+        reflectometry data in the file. It's contained in a list because each
+        of the datasets may have a different number of columns.
+    """
+
     with _possibly_open_file(file, "r") as fi:
         header = []
-        for line in fi.readlines():
+
+        # variables to keep track of where the datasets are
+        ds_lines = []
+        in_ds = False
+        start_line = -1
+        end_line = -1
+
+        for i, line in enumerate(fi.readlines()):
             if not line.startswith("#"):
+                if not in_ds:
+                    # you're in the first line of a dataset
+                    in_ds = True
+                    start_line = i
                 continue
+            if in_ds:
+                # you've reached the first line after a comment line
+                end_line = i - 1
+                ds_lines.append((start_line, end_line))
+                start_line = end_line = -1
+                in_ds = False
+
             header.append(line[1:])
-        return "".join(header)
+
+        # finished reading the file, have to append the last dataset
+        ds_lines.append((start_line, i))
+
+        data = []
+        for ds_line in ds_lines:
+            fi.seek(0, 0)
+            start, end = ds_line
+            arr = np.loadtxt(fi, skiprows=start, max_rows=end - start + 1)
+            data.append(arr)
+
+        return "".join(header), data
 
 
 def _validate_header(h: str):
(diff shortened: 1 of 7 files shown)


andyfaff:
@arm61 @jochenstahn @bmaranville @aglavic @maxskoda This PR is in a reasonable condition now. However, there are two concerns.

1. The file ingestion code in `Orso.from_file` is reasonably convoluted. This is because there is no easy way to recreate the `Orso` object from the dict that's created by `yaml.safe_load`. If we added yaml tags to the ORSO file then object reconstruction might be a little easier, e.g. https://matthewpburruss.com/post/yaml/. Unfortunately adding tags doesn't look the best. 
Artur suggested an approach in the [`fromdict` branch](https://github.com/reflectivity/orsopy/tree/fromdict/orsopy), but when I tried investigating this it would get stuck on attributes that weren't simply typed. e.g. `owner: Person` would work, but `owner: Union[Person, List[Person], str]` would not.

2. This point is more concerning. At the moment the ORSO file specification is supposed to allow multiple datasets by separating them by extra `data_source` yaml sections. However, the yaml specification does not allow duplicate tags in a single document, see e.g. https://github.com/yaml/pyyaml/issues/165. In other words:
```
# unacceptable trying to put two data_sources in a YAML file
data_source:
    owner:
        name: Andrew
        affiliation: Sydney
data_source:
    owner:
        name: Andrew2
        affiliation: Denmark

# acceptable way to have two data_sources in a YAML file is to have a list.
data_source:
- owner:
    name: Andrew
    affiliation: Sydney
- owner:
    name: Andrew2
    affiliation: Denmark
```
Whilst we could use extra file manipulations to work around this, the file considered as a whole would still not be valid YAML. One way to get around this would be to have one of the following layouts (where the `data_set` attribute is moved from `Orso` to `Orso.data_source`):
```
data_source:
- owner:
       name: Andrew
       affiliation: Sydney
   data_set: spin-up
<numerical data for the first dataset>
- owner:
       name: Andrew2
       affiliation: Denmark
   data_set: spin-down
<numerical data for the second dataset>

OR

data_source:
- owner:
       name: Andrew
       affiliation: Sydney
   data_set: spin-up
- owner:
       name: Andrew2
       affiliation: Denmark
   data_set: spin-down
<numerical data for the first dataset>
# separator to denote next dataset, e.g. # +++++++
<numerical data for the second dataset>
```

Either of these suggestions would require a small re-write of the specification regarding the separation of successive datasets, etc.

bmaranville:
I think the idea was never to treat the entire data file as valid YAML.  There was always going to be a separate step for identifying and extracting the header(s).

andyfaff:
I agree @bmaranville. According to the current specification the YAML header(s) text is prefixed by the `#` character and the numerical data is written without that character, i.e.:

```
#creator:
#  name: Andrew
0.01, 1.0, 0.001, 0.0005
```
As such the first step in reading the file is to iterate through the lines to extract the header text, stripping off that leading character, leaving one with text that represents a YAML fragment. That fragment can then be loaded via `yaml.safe_load` to give a dictionary. From my viewpoint *that YAML fragment should be loadable as-is, without further processing* (it doesn't take too much effort to extract the numerical arrays).
Unfortunately the current specification has successive arrays, each separated by `data_source` tags.

```
## INVALID YAML because data_source and data_set appear repeatedly at the same level
#data_source:
#  owner:
#    andrew2
#data_set: 0
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005
#data_source:
#  owner:
#    andrew2
#data_set: 1
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005
```
According to the YAML specification this is illegal, one should only see a given tag once. The proper way of doing this is to create an array:

```
## VALID YAML, option 1
#data_set:
#- 0
#- 1
#data_source:
#- owner:
#     andrew
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005
#- owner:
#     andrew2
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005

-OR-

## VALID YAML, option 2
#data_set:
#- 0
#- 1
#data_source:
#- owner:
#     andrew
#- owner:
#     andrew2
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005
## SOME SORT OF SEPARATOR
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005
```

Either of those two 'valid' options don't align with the specification at the moment. Therefore, we can either make a few small changes to the specification, or go through a lot of pain trying to massage text into something that's valid YAML. I believe the former is much much easier. The changes are:

1.  Agree on option 1 or 2. If we go with option 2 then we need to define the separator between the two arrays.
2. With option 1 or 2 above the `data_set` attribute is a separate list, and not directly associated with a `data_source`.
I suggest making `data_set` part of the `data_source`:
```
## VALID YAML, option 1modified
#data_source:
#- data_set: 0
#  owner:
#    andrew
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005
#- data_set: 1
#  owner:
#    andrew2
0.01, 1.0, 0.001, 0.0005
0.01, 1.0, 0.001, 0.0005
```
Either of the options is easy to implement in code, and is easy to change in the specification.

PR #20, opened 2021-08-17
Fromdict with load/safe

aglavic:
Following the various discussions I've finished implementing the creation from dictionary and made a simple load and safe function that should create valid single+multi dataset files.

There is an example file 'writetest.py' that writes and loads a single .ort file.


aglavic:
> I might be missing something (I haven't been keeping upto date with changes this week and last, but what is the rationale for removing the `orso_optionals` in place of just getting the optional attributes? [#2 (comment)](https://github.com/reflectivity/orsopy/pull/2#issuecomment-874759776) In this PR @bmaranville and I determined the `orso_optionals` list necessary.

The point of removing that attribute is to make the classes more readable and, at the same time, make automatically sure that the list contains all Optional arguments. This is now implemented in a Metaclass that also adds a comment field to all items.

aglavic:
@bmaranville 
Have a look at the implementation. It's inspecting the field type and only sets it as optional if type(None) in get_args(field.type), which is equivalent with a type hint of Union[somthing..., None] = Optionsal[somthing...].

orsopy/fileio/base.py
```diff
@@ -34,17 +37,113 @@ def __datetime_representer(dumper, data):
...
+                if updt is not None:
+                    # convert to dataclass instance
+                    setattr(self, fld.name, updt)
+                else:
+                    raise ValueError(f"No suitable conversion found for {fld.type} with value {attr}")
+        if hasattr(self, 'unit'):
             self._check_unit(self.unit)
 
+    @staticmethod
+    def _resolve_type(hint, item):
```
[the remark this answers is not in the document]
andyfaff: If I understand correctly this method recursively converts attributes that are in json dict form to Header objects?

_resolve_type(Person, {"name": "Andrew", "affiliation": "somewhere"}) --> Person

orsopy/fileio/base.py
```diff
@@ -34,17 +37,113 @@ def __datetime_representer(dumper, data):
...
+            if isinstance(item, hint):
+                return item
+            if issubclass(hint, Header):
+                # convert to dataclass instance
+                try:
+                    return hint(**item)
+                except (ValueError, TypeError):
+                    return None
+            else:
+                # convert to type
```
andyfaff, line 101: at what points does this happen?

orsopy/fileio/base.py
```diff
@@ -34,17 +37,113 @@ def __datetime_representer(dumper, data):
...
+                try:
+                    return hint(item)
+                except (ValueError, TypeError):
+                    return None
+        else:
+            # the hint is a combined type (Union/List etc.)
+            hbase = get_origin(hint)
+            if hbase is list:
```
andyfaff, line 109: What happens if hbase is a tuple? The code may need to be:
```
if issubclass(hbase, collections.abc.Sequence):
```

orsopy/fileio/base.py
```diff
@@ -34,17 +37,113 @@ def __datetime_representer(dumper, data):
...
+                t0 = get_args(hint)[0]
+                if type(item) is list:
```
andyfaff, line 111: `if issubclass(type(item), collections.abc.Sequence):`

orsopy/fileio/orso.py
```diff
@@ -4,45 +4,180 @@
...
+            np.savetxt(f, dsi.data.T, header=hi, fmt='%-22.16e')
+
+
+def read_data(text_data):
+    # read the data from the text, faster then numpy loadtxt with StringIO
+    data = [li.split() for li in text_data.strip().splitlines() if not li.startswith('#')]
+    return np.array(data, dtype=float).T
+
+
+def load(fname: Union[TextIO, str]):
```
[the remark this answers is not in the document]
andyfaff: Also, see argument above. Why is a separate `OrsoDataset` necessary, why not place numerical data into `Orso` objects?:

```
def load_orso(fname: Union[TextIO, str]) -> List[Orso]
```

orsopy/fileio/orso.py
```diff
@@ -4,45 +4,180 @@
...
-    _data_source = DataSource(
-        _person, _experiment, Sample(None), _measurement
-    )
-
-    _reduction = Reduction(Software(None, None, None), None, None, None)
-    return Orso(_creator, _data_source, _reduction, None, None)
+    def from_difference(self, other_dict):
+        # recreate info from difference dictionary
+        output = self.to_dict()
+        self._update_dict(output, other_dict)
```
andyfaff, line 48: use `base._nested_update`

andyfaff (changes requested):
Hey @aglavic, several comments here. I've tried to be rigourous. I guess the most significant are:

1. Is it necessary to have a HeaderMeta class? It's only use seems to be creating `_orso_optionals` from the type hints. I reckon it's easier to have them explicitly listed in the class definition themselves. The metaclass programming seems overly complicated for what it offers.
2. Why are there separate `Orso` and `OrsoDataset` classes? The latter only has an extra data attribute. I would've thought it more logical to just add the `data` attribute to `Orso` instead.
3. There are several staticmethod dict manipulations in Orso, they should probably be standalone dict utility functions.

writetest.py
```diff
@@ -0,0 +1,66 @@
...
+from datetime import datetime
+from orsopy import fileio
+
+
+def main():
+    info=fileio.Orso.empty()
+    info2=fileio.Orso.empty()
+    data=np.array([np.arange(100), np.arange(100), np.arange(100)])
+    info.columns=[
+        fileio.Column('q', '1/A'),
```
andyfaff, line 11: I believe this should be "Qz", "1/angstrom"

writetest.py
```diff
@@ -0,0 +1,66 @@
...
+                date=datetime.now(),
+                probe='x-rays',
+            ),
+            owner=fileio.Person('someone', 'important'),
+            measurement=fileio.Measurement(
+                instrument_settings=fileio.InstrumentSettings(
+                    incident_angle=fileio.Value(13.4, 'deg'),
+                    wavelength=fileio.Value(5.34, 'A')
+                ),
+                data_files=['abc', 'def', 'ghi'],
```
andyfaff, line 49: How about if one of the items is a `File` object?

orsopy/fileio/orso.py
```diff
@@ -4,45 +4,180 @@
...
-    _experiment = Experiment(None, None, None, None)
-    _instrument_settings = InstrumentSettings(None, None)
-    _measurement = Measurement(_instrument_settings, None)
-    _data_source = DataSource(
-        _person, _experiment, Sample(None), _measurement
-    )
-
-    _reduction = Reduction(Software(None, None, None), None, None, None)
-    return Orso(_creator, _data_source, _reduction, None, None)
+    def from_difference(self, other_dict):
```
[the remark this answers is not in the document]
andyfaff: I'll take this one back

andyfaff:
@arture, I rebased and force-pushed to remove the merge conflicts. I have a backup from before the force push on my local computer (from_dict_orsopy_backup)

orsopy/fileio/base.py
```diff
@@ -34,17 +37,113 @@ def __datetime_representer(dumper, data):
 yaml.add_representer(datetime.datetime, __datetime_representer)
 
 
-class Header:
+class HeaderMeta(type):
```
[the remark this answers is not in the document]
andyfaff: Having gone through the code more closely I now understand how the Metaclass works to construct the _orso_optionals from the typing. However, I do wonder how robust this will be in the long run.

andyfaff:
Even though there's test failures I've merged this PR, because the history is quite complex, thanks @aglavic. I'll sort out the fails today.

orsopy/fileio/orso.py
```diff
@@ -4,45 +4,180 @@
...
+class OrsoDataset:
+    info: Orso
+    data: np.ndarray
+
+    def __post_init__(self):
+        if self.data.shape[0] != len(self.info.columns):
+            raise ValueError("Data has to have the same number of columns as header")
+
+    def header(self):
+        out = self.info.to_yaml()
```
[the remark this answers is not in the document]
andyfaff: probably not, because `diff_header` is similar and it needs to be a method.

PR #23, opened 2021-08-18
ENH: add a LICENSE file

andyfaff:
At the moment setup.py states that orsopy uses the MIT licence. This PR includes that licence in the source repository. MIT is a permissive licence enabling it to be used in closed source, open source projects.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020, Orsopy developers, ORSO organisation
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


arm61:
There was some discussion previously on what license we should use so I removed the file. Then the discussion never reached a conclusion. 

Personally I like MIT

andyfaff:
A non permissive licence will be cause issues for various packages. e.g. if it is (L)GPL this will cause issues including it in other bundles (e.g. a standalone exe).

I guess I'll leave the PR open until we've decided. However, it's not the best to leave a repository without a licence. In the near future we may make a PyPI release, and we'll definitely need one by then.

arm61:
Agreed, like I say I like the simplicity of MIT. 

@aglavic @bmaranville any issues with MIT?

bmaranville:
I'm ok with MIT - on some level I'd like to encourage the companies to open-source their tools, but I think getting them to engage is more important.

PR #29, opened 2021-08-24
ENH: user_data

andyfaff:
@aglavic, did you have something like this in mind for `user_data`? Note that you can't have a `**kwds` parameter for a dataclass.

Also, when I was writing a test I found that there was a problem with saving/reloading an ORT file with a single dataset. Currently the specification doesn't mandate that `data_set` be present in a file when there is only a single dataset:

> If there are multiple data sets in one file, the first one can be given an identifier with the optional line:

The existing code doesn't save the `data_set` attribute when [`data_set == 0`](https://github.com/reflectivity/orsopy/blob/c5c7cdc68660ff927ce1b95b490af5625e5d4145/orsopy/fileio/base.py#L173)
This is a problem when reconstructing the `Orso` object because the constructor requires `data_set` as a positional parameter, but it's not present in the loaded dict (`Orso(**dct)`). To get around this I made `data_set` a keyword argument with a default of None, moving it after the positional args. I also typed it as `Optional` because it's not needed when working with a single dataset.

Alternatively we can make it mandatory in the specification that `data_set` is provided for each dataset, and make sure it's written in a file. In which case we can revert the changes in this PR, removing the  [`data_set == 0` check](https://github.com/reflectivity/orsopy/blob/c5c7cdc68660ff927ce1b95b490af5625e5d4145/orsopy/fileio/base.py#L173).

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -82,7 +82,14 @@ class Header(metaclass=HeaderMeta):
                     # convert to dataclass instance
                     setattr(self, fld.name, updt)
                 else:
-                    raise ValueError(f"No suitable conversion found for {fld.type} with value {attr}")
+                    warnings.warn(
+                        f"No suitable conversion found for {fld.name}, "
+                        f"{fld.type} with value {attr}, "
+                        "setting attribute with raw value from ORSO file",
+                        RuntimeWarning
+                    )
+                    setattr(self, fld.name, attr)
+                    # raise ValueError(f"No suitable conversion found for {fld.type} with value {attr}")
         if hasattr(self, 'unit'):
             self._check_unit(self.unit)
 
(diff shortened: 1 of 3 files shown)


andyfaff:
I've only added `user_data` to the `Orso` class. Does it need to be added to every class, via `HeaderMeta`?

aglavic:
@andyfaff 

I had a look at your changes. Yes, it makes sense to make data_set optional, we just have to make sure on the export that the columns come last.

I had a look at the Dataclasses framework and I don't think it is a big problem to allow user data as additional keywords. The only thing required is to replace the Dataclass __init__ with a custom one. I'll make a quick implementation and create a pull request before our meeting later.

andyfaff:
I'm not sure we'd want to replace the default `__init__`, do you mean `__post_init__`?

PR #38, opened 2021-09-02
MAINT: datetime loaded as str

andyfaff:


andyfaff:
@bmaranville this PR loads datetimes as strings, following your snippet. These are then converted to datetime objects at a later stage. The jsonschema validation then doesn't need modifying.
I've also modified the test_example.ort file to include +01:00 strings on the datetimes (apart from the range)

PR #44, opened 2021-09-24
Implement user_data from custon keyword arguments

aglavic:
This implementation allows the users to add additional "groups" to the header by supplying keyword arguments to Orso object.

The advantage over a user_data attribute are, that the recreation from the header works directly and that it will be forward/backward compatible if we adapt a new group to the standard.
We'll have to discuss if this should be pushed upward to Header to allow user data in any block, but that might make things less clean.

@andyfaff have a look if this works as you intended.

### DIFF
--- a/orsopy/fileio/orso.py
+++ b/orsopy/fileio/orso.py
@@ -20,7 +20,7 @@ ORSO_DESIGNATE = (f"# ORSO reflectivity data file | {ORSO_VERSION} standard "
                   "| YAML encoding | https://www.reflectometry.org/")
 
 
-@dataclass(repr=False)
+@dataclass(repr=False, init=False)
 class Orso(Header):
     """
     The Orso object collects the necessary metadata.
@@ -32,6 +32,28 @@ class Orso(Header):
     data_set: Optional[Union[int, str]] = None
 
     __repr__ = Header._staggered_repr
+
+    def __init__(self, creator: Creator, data_source: DataSource, reduction: Reduction,
+                 columns: List[Column], data_set: Optional[Union[int, str]]=None, **user_data):
+        self.creator=creator
+        self.data_source=data_source
+        self.reduction=reduction
+        self.columns=columns
+        self.data_set=data_set
+        self.__post_init__()
+        # additional keywords used to add fields to the file header
+        # some recreation does not work when using the attribute directly so it's wrapped in a property
+        self._user_data=user_data
+
+    @property
+    def user_data(self):
+        return self._user_data
+
+    @user_data.setter
+    def user_data(self, value):
+        if not type(value) is dict:
+            raise ValueError("user_data has to be a dictionary")
+        self._user_data=value
 
     def column_header(self) -> str:
         """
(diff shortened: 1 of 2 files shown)


andyfaff:
Instead of making the dataclass in `__new__`, would it be easier to use the `@dataclass` decorator on the `Header` instead?

@aglavic on a workflow note, it might make more sense for you to fork the `orsopy` repository to your own GH account, and make future feature branches there. It makes the list of branches held on this repository much shorter, and removes the possibility of different users pushing to the same branch by mistake.

aglavic:
The modifications are ready to merge, @andyfaff  and @arm61, please have a look.

@andyfaff about the branches: The forking seems unnecessary extra effort on both the remote and local repository end. Isn't the point of a GIT repository to work on code together.
I don't see the issue with extra pull-request branches, as they are getting deleted after the PR is closed, right?

andyfaff:
Great, I'll take a look soon.

The workflow I suggested is used by other mature projects, e.g. [scipy/numpy](https://numpy.org/doc/stable/dev/index.html). There's still only one feature branch, it's just done on contributors forks. The workflow makes it easier for projects as they grow.

aglavic:
OK, I can try next time. I never forked a project so I'll have to learn how the workflow is done correctly.

I think you are a bit optimistic to think that orsopy will grow to a large project with many contributors like numpy, though ;-)

arm61:
This looks good to me.

arm61:
@andyfaff shall we merge this?

andyfaff:
Sorry, on leave this week.

arm61:
@aglavic if you rebase to `main` the tests should pass again.

aglavic:
@arm61 I'll wait until it's decided if this PR is accepted to avoid double effort. Although I don't understand why this branch can fail it's tests due to changes to another branch...

arm61:
The changes are upstream in the `jsonschema` package, which made a new release on the 30th September. I was having the tests fail on a branch I created that didn't make sense (cause I had only touched the documentation) so I reran the tests on this branch, which used the new version of `jsonschema==4.0.1` instead of the version your tests has originally run on `3.2.0`. Rebasing will pull in the pinning to this version that I implemented in #50.

aglavic:
@andyfaff can you review this PR so we can close it before it diverges too far from what @arm61 is doing?

andyfaff:
Yes, sorry, I was busy this week. I'll do it earlier next week.

aglavic:
@andyfaff 
I think auto-formatting is fine in cases like spaces around operators etc. I'm not so sure about line breaks and continuation indents, but I won't object if you guys prefer that.

andyfaff:
If we use black autoformatter we can specify line length, etc. We can do it in a precommit hook to save frustration.

arm61:
There are benefits to having a standard style even if the style isn't to everyones preference (in particular for projects with contributors that aren't necessarily in close contact). `black` works for me cause it is pretty comprehensive, even if it makes things look a bit weird.

PR #50, opened 2021-10-06
pin jsonschema to 3.2.0

arm61:
jsonschema released a 4.0.1 update that appears to break the schema testing. I have pinned to 3.2.0 (the previous release) until we can investigate further.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -11,5 +11,5 @@ numpy>=1.19.0
 
 pytest==6.2.4
 pyyaml>=5.4.1
-jsonschema
+jsonschema==3.2.0
 typing_extensions


requirements_dev.txt
```diff
@@ -11,5 +11,5 @@ numpy>=1.19.0
 
 pytest==6.2.4
 pyyaml>=5.4.1
-jsonschema
+jsonschema==3.2.0
```
andyfaff, line 14: Would have been better as `<4`?

bmaranville: How are you testing (how did you find out it was breaking?)

bmaranville: https://github.com/Julian/jsonschema/issues/847

PR #54, opened 2021-10-08
PyPI integration

aglavic:
Preparing the automatic packaging and upload to PyPI when a commit is version tagged.


arm61:
Dunno if this will help, but I implemented this in another project so that when a release is made on github a new release is shipped to pypi
https://github.com/bjmorgan/kinisi/blob/master/.github/workflows/release.yml

PR #56, opened 2021-10-08
Integrated schema gen

bmaranville:
Move all schema generation to use orsopy.fileio classes.


bmaranville:
I need to rebase on end of main before this PR - will do again.

PR #57, opened 2021-10-08
Integrated schema gen

bmaranville:
Generate schema using the same classes being used to read/write headers.

The generated schema is no longer validating the example files, so once that's figured out this can be merged.


andyfaff:
Not sure if an 'empty' header should validate either.

bmaranville:
I have a working version of this now, based on the metaclass2decorator branch.  Once that PR is merged again I will make another (new) pull request for this feature.

PR #59, opened 2021-10-19
Backport to python 3.6 and 3.7

aglavic:
Trying to test .ort creation in our PSI reduction I found limited access to newer python versions. As the requirement for python >=3.8 comes from the typing inspection we do in the Header I backported these to work with the pip installed typing_extensions and dataclasses.

I'm not sure, if we should change the official requirement for the PyPI release, though.

### DIFF
--- a/tests/test_fileio/test_orso.py
+++ b/tests/test_fileio/test_orso.py
@@ -115,6 +115,10 @@ class TestOrso(unittest.TestCase):
         value = Orso(ds, redn, cols, '1')
         assert value.data_set == '1'
 
+    def test_repr(self):
+        ds = fileio.Orso.empty()
+        repr(ds)
+
     def test_write_read(self):
         # test write and read of multiple datasets
         info = fileio.Orso.empty()


arm61:
I have no problems with the code. I think the philosophical question about what it on PyPI needs broader discussion. Thoughts @bmaranville @andyfaff?

bmaranville:
I tend to agree that supporting older versions of python is worth doing especially if it doesn't cost very much.  For distributed, packaged binaries etc. of course use the current version of python.  We have plenty of installed hardware running old _but stable_ linux distros, for which we are not the sysadmins.

andyfaff:
The code in itself looks fine.

From a philosophical point of view:
- supporting older Python versions does require time (even if it's on the lower end of things).
- it prevents newer language features being used.
- From December Python 3.6 no longer receives security updates, it's end-of-life. This means CI engines will stop directly supporting it. Supporting Python versions that no longer receive security updates is sending the wrong message. I'd argue that for facilities to use out of date Python versions is a security risk.
- Stability and back compatibility is important.  That long term stability is vital for servers (hence LTS versions of Linux), but is mainly needed for system libraries. It's often possible for more recent versions of Python to be installed in user accounts, especially if conda is accessible. If the machine is so locked down that installing new software is not easily achievable, then one will have just a hard a time installing orsopy itself. It should be as easy to install a new Python version as any package (hence the popularity of conda).
- What would be our policy for supporting older Python versions? Python/numpy/scipy/etc have developed policies for this; for example the master branch of [numpy has already dropped 3.7](https://github.com/numpy/numpy/blob/main/setup.py#L94). Should we be at the mercy of the slowest facility(ies) to use a more modern version of Python? When would we drop 3.6?

If we do merge the PR I'd like for us to have discussed the last point prior to that.

PR #60, opened 2021-10-20
Allow spaces between datasets

aglavic:
Jochen pointed out that the specification allows for empty lines in/between datasets. 

This implements an optional spacer string when saving/loading.


orsopy/fileio/orso.py
```diff
@@ -182,7 +182,8 @@ def __eq__(self, other: 'OrsoDataset'):
 def save_orso(
         datasets: List[OrsoDataset],
         fname: Union[TextIO, str],
-        comment: Optional[str] = None) -> None:
+        comment: Optional[str] = None,
+        data_separator: str = '') -> None:
```
andyfaff, line 186: `data_separator` needs to be added to the docstring. Are there any separators which would cause havoc? Should there be a restricted set allowed, `{'\r', '\r\n', '', '\n')`? If there isn't a restricted set then it would be easy to write a non-conformant dataset, e.g. if someone provided `'abc'` then the array creation would fail when it tried to create an array.

orsopy/fileio/base.py
```diff
@@ -501,7 +501,9 @@ def _read_header_data(file: Union[TextIO, str], validate: bool = False) -> Tuple
 
         for i, line in enumerate(fi.readlines()):
             if not line.startswith("#"):
-                _ds_lines.append(line)
+                # ignore empty lines
+                if line.strip() != '':
+                    _ds_lines.append(line)
```
andyfaff, line 506: Would this be better:
```
if not line.startswith("#"):
    _ds_lines.append(line.strip())
```
Alternatively the more Pythonic version of the if statements is:
```
if not line.startswith("#") and line.strip():
    _ds_lines.append(line)
```

PR #61, opened 2021-10-29
Replace Meta Class by decorator of same functionality

aglavic:
I think some of you were not so happy with the Meta Class solution for the creation of the dataclasses and it turns out that some static analyzers like the one used by PyCharm can't handle this well and generate false positives for wrong parameters when creating the objects.

This is an alternative that replaces it with a decorator solution of equivalent functionality. This way we also follow the same logic as the build-in dataclass.


bmaranville:
I am very concerned - this whole pull request vanished from the history.  This should not happen - EVER.

PR #62, opened 2021-10-29
Include slddb sub-package in release

aglavic:
The slddb website allows access via API requests and includes a python package that implements this access.

This change clones the package during PyPI build into orsopy so it can be distributed together.
Example how to use it:
```python
from orsopy.slddb import api, constants

results=api.search(formula='Co')
m=api.material(results[0]['ID'])
print(f"neutron SLD is {m.rho_n}")
print(f"Cu k-alpha SLD is {m.rho_of_E(constants.Cu_kalpha)}")

p=api.bio_blender('aaba', molecule='protein')
print(f'Protein match point is {100*p.match_exchange(D_fraction=0.0, exchange=0.9):.2f}% for 90% exchange')
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,10 @@ jobs:
         sudo apt update
         sudo apt install python3 python3-setuptools python3-pip python3-yaml
         pip3 install build
+    - name: Add slddb package
+      run: |
+        git clone https://github.com/reflectivity/slddb.git slddb_source
+        mv slddb_source/slddb orsopy/
     - name: Build PyPI package
       run: |
         python3 -m build
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include LICENSE
 include README.rst
 
 recursive-include tests *
+recursive-include orsopy/slddb/element_table/nabs_geant4 *.npz
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,9 +23,10 @@ classifiers =
 
 [options]
 python_requires = >=3.8
-packages =
-    orsopy
-    orsopy.fileio
+packages = find:
 install_requires =
     pyyaml>=5.4.1
 include_package_data = True
+
+[options.package_data]
+orsopy.slddb.element_table = nabs_geant4/*.npz


bmaranville:
What is the motivation for distributing these packages together?  Is slddb needed for the operation of orsopy libraries?

andyfaff:
@bmaranville, I'm guessing if the code output of ORSO was going to go anywhere one would expect it to be in `orsopy`; the fileio of the file-format working group went into `orsopy.fileio`. 
From the changes here it looks like the slddb code would be included when a wheel is built in a release step. The fileio code would not depend on the slddb code being included or not.
I'm not sure of the utility of the database being included - a new user would probably refer to a web interface.

arm61:
Yeah, as @andyfaff says, IMHO the long term aim for `orsopy` is to be a Python package for the code ORSO produces (although if there is strong dissent then we can have a discussion on it).  
I personally like the idea of including the database in the wheel release (or even bringing the code in as `orsopy.slddb`) as I can reduce the number of dependencies (i.e. my analysis software will use `orsopy.fileio` for reading files and `orsopy.slddb` for finding scattering length densities for different materials.

aglavic:
@bmaranville yes, the idea is as stated by @andyfaff that `orsopy` is a package distributing all ORSO developed code, where this makes sense. The `slddb` sub-package does not require any additional dependencies. It could be included directly in this repository but I wanted to avoid two parallel developments as the sub-package will be the same as in the slddb repo.

@andyfaff it is not adding the database itself, just the package that provides an interface for query and calculation. It implements the webapi in python, allows offline SLD calculation and creates a local copy of the database (<once a day) for off-line query.
The use case I was thinking about is not for end-users but for other programs that want to integrate database query. I have something like this in GenX for some crystallographic databases.

aglavic:
@bmaranville 
I don't think PyPI allows this kind of sub-packaging, one would require separate packages.

The advantage I see is, that users find all our stuff in one place (as with scipy) and might explore new options. We don't have many requirements and the package will stay pretty small, anyway. 

Concerning the versioning: I think we already have different versions for the file format itself. That's not an issue per-se, we could just release a new orsopy sub-version when ever a sub-package get's updated.

bmaranville:
I did mean separate packages, just with similar names (adding orsopy. at the beginning of each one).  I think if we want them to be one package, we should maintain them in a single repository.

arm61:
My vote is to bring `slddb` fully into `orsopy` at the repository level.

aglavic:
@arm61 I'm fine with this, if I don't need to impose the same formatting rules onto slddb. 

Although this means a bit of extra manual work for an update, I don't think the package will be updated so often that it's unfeasible.

@all It seems we have quite a variety of opinions on this matter. What would be the right format to define a way forward? A quick video meeting in small circle or the broader ORSO community?

arm61:
Curious about your opposition to the formatting. Is it all specified formatting you are opposed to or just the 80 char line length? 

The additional benefit to bringing `slddb` into `orsopy` is that there are more of use working on `orsopy` so there is the chance for code review/improvements to documentation/etc. 

I think that a `orsopy` developers meeting should suffice for this discussion. Should I send an email round with the prerequisite doodle poll?

aglavic:
I'm not opposed to formatting guidelines, in general. It's just the strict PEP8 enforcement. There are some PEP8 guidelines that I personally don't like (e.g. spaces around everything) and others where I just find it annoying to enforce them in every instance.

E.g. in some formulas I use different number of indents to make them more clear and I don't like it so much to be forced to change all that code, that is pretty readable, to follow a style guideline like a law

aglavic:
To keep things clean, I'll create a new PR for including the package.

PR #63, opened 2021-12-02
Meeting minutes form 2021-11-30

arm61:
Please let me know if there is anything you disagree with or I have forgotten.

### DIFF
--- a/docs/developer_minutes/meeting_2021_11_30.rst
+++ b/docs/developer_minutes/meeting_2021_11_30.rst
@@ -26,17 +26,44 @@ Minutes
 
   - AN introduced the motivation for collaborative projects towards having a common code style (via formatting rules). 
   - It was broadly agreed that a common style is beneficial, however, AG would like to see it applied in a soft fashion, i.e. a pull request does not fail for the wrong style (this could be offputting to new contributors). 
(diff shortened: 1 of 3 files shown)


aglavic (approved):
All fine with me

bmaranville (approved):
These all look good.

bmaranville:
I don't know where to put this suggestion, but I would like to add to our contribution guidelines a hard rule: "no force pushes".  Clearly there will be extraordinary circumstances where this will have to be violated (accidental committing of sensitive material, etc) but in general I think the risk is too great.

andyfaff:
I agree with the no force pushes suggestion.

The only time I would do a force push would be to a feature branch on my fork. When the feature branch arrives in a PR the changes are then obvious. Do you know how much code was mashed?

bmaranville:
I was able to recover the commits like this:
```bash
git fetch origin pull/61/head:metaclass2decorator
git checkout metaclass2decorator
```
as suggested in https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

acaruana2009 (approved):

arm61:
I have added something about `force push`. But it looks to me like there is enough approvals, so I will go ahead with it.

PR #64, opened 2021-12-03
Guidelines how to autoformat the code

aglavic:
As preparation to transfer slddb to orsopy I have looked into the usage of black and ran it on the project. In the same spirit I found that isort would also be a good addition for us to keep the import section tidy and equivalent for the whole package.

I have therefore added a description how to run this to the contributing section of the documentation. The code is also updated to what it would looks like after running these commands.

One additional thing I changed might need some discussion here. I removed the author comments that have been in several files for the following reasons:

1. Most files are touched by many of us, so they don't really have an Author in that sense.
2. We do have author information in a separate file now so that everyone get's the credit they deserve.
3. The git repository itself is a better indication of who contributed what to the project (and whom to ask about it, potentially)

Any comments are welcome. Black doesn't have any configuration parameters, so we will have to live with the look of the results, but for me the changes mostly looked reasonable.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,31 @@
+name: Code Format Check
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.9]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8
+          if [ -f requirements_dev.txt ]; then pip install -r requirements_dev.txt; fi
+
+      - name: Lint
+        run: |
+          flake8 --max-line-length=120 --ignore=F401,W503 --count --show-source --statistics orsopy
+          flake8 --max-line-length=120 --ignore=F401,W503 --count --show-source --statistics tests
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -76,12 +76,12 @@ Ready to contribute? Here's how to set up `orsopy` for local development.
 
    Now you can make your changes locally.
 
-5. When you're done making changes, check that your changes pass flake8 and the
-   tests, including testing other Python versions with tox::
+5. When you're done making changes, auto format the code and check that your changes pass the unit
+   tests::
 
-    $ flake8 orsopy tests
+    $ black -l 120 orsopy tests
+    $ isort -l 120 --lbt 1 orsopy tests
     $ python setup.py test or pytest
-    $ tox
 
    To get flake8 and tox, just pip install them into your virtualenv.
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -81,7 +81,7 @@ Ready to contribute? Here's how to set up `orsopy` for local development.
 
     $ black -l 120 orsopy tests
     $ isort -l 120 --lbt 1 orsopy tests
-    $ python setup.py test or pytest
+    $ pytest
 
    To get flake8 and tox, just pip install them into your virtualenv.
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -77,10 +77,11 @@ Ready to contribute? Here's how to set up `orsopy` for local development.
    Now you can make your changes locally.
 
 5. When you're done making changes, auto format the code and check that your changes pass the unit
-   tests::
+   tests and confirms to PEP 8::
 
     $ black -l 120 orsopy tests
     $ isort -l 120 --lbt 1 orsopy tests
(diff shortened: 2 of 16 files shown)


aglavic:
And please ignore the coverage decrease, that's just the result of the change of the number of lines in the same untested code sequence.

CONTRIBUTING.rst
```diff
@@ -76,12 +76,12 @@ Ready to contribute? Here's how to set up `orsopy` for local development.
...
 
-5. When you're done making changes, check that your changes pass flake8 and the
-   tests, including testing other Python versions with tox::
+5. When you're done making changes, auto format the code and check that your changes pass the unit
+   tests::
 
-    $ flake8 orsopy tests
+    $ black -l 120 orsopy tests
+    $ isort -l 120 --lbt 1 orsopy tests
     $ python setup.py test or pytest
```
arm61, line 84: Maybe we just go for (I know this is code that I added in the first place)
```suggestion
    $ pytest
```

orsopy/__init__.py
```diff
@@ -1,5 +1,5 @@
 """Top-level package for orsopy."""
 
 __author__ = """Andrew R. McCluskey"""
-__email__ = 'andrew.mccluskey@ess.eu'
-__version__ = '0.0.3'
+__email__ = "andrew.mccluskey@ess.eu"
```
arm61, line 4: We should maybe remove this since it is becoming more and more redundant with more authors

arm61 (approved):
Some small thoughts.

orsopy/__init__.py
aglavic: Agreed, better to keep this on github/documentation files than the package directly.
I checked that the setup does not use these variables.

bmaranville (approved):
Having an agreed-upon formatting will be very helpful - thanks for all the work, @aglavic !

arm61:
How come the `force push` is needed?

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8
+          if [ -f requirements_dev.txt ]; then pip install -r requirements_dev.txt; fi
+
+      - name: Lint
+        run: |
+          flake8 --max-line-length=120 --ignore=F401,W503 --count --show-source --statistics orsopy
+          flake8 --max-line-length=120 --ignore=F401,W503 --count --show-source --statistics tests
```
andyfaff, line 31: I think that if we're suggesting (in CONTRIBUTING.rst) that people call `black` before code is committed, then we should also add the same black test here.

andyfaff, line 31: `black -l 120 orsopy tests`

aglavic: @andyfaff I don't think this would be useful. The action would change the code that is tested by flake8 while not changing it in the repository. An auto-commit in the action, on the other hand, could lead to confusion.

aglavic: See above

arm61: Yeah I think this would just change the code to black formatting and then not do anything... I am not sure how black works on CI though so it might fail (if it works like `yapf`

andyfaff: Sorry, it should have the --check flag

andyfaff: Then it sees if the formatting is correct.

bmaranville: I agree - check but don't change the code (commit or push anything) in CI.  This all seems ok to me.

aglavic: The check will fail, even if the user followed the formatting guidelines as `isort `is run after `black`. This means that long `from ... import ....` statements will be split up on many lines by black. This was one of the reasons I liked running isort, it not only creates the same import orders for all files it also makes the format more readable than an import statement that runs of 10-20 lines.

95% of what black does is making code compatible with PEP 8, so code that is not formatted by black will likely fail `flake8`. I would therefore see the test as redundant. The possible existing formatting errors from someone not running `black` but passing `flake8` can only be minor.

aglavic: Here is how blacks diff would look like on the current branch (from isort version back to black version):
```diff
--- orsopy\fileio\base.py       2021-12-06 20:07:25.239734 +0000
+++ orsopy\fileio\base.py       2021-12-09 07:57:32.837696 +0000
@@ -10,12 +10,25 @@
 import warnings

 from collections.abc import Mapping
 from contextlib import contextmanager
 from copy import deepcopy
-from dataclasses import (_FIELD, _FIELD_INITVAR, _FIELDS, _HAS_DEFAULT_FACTORY, _POST_INIT_NAME, MISSING, _create_fn,
-                         _field_init, _init_param, _set_new_attribute, dataclass, field, fields)
+from dataclasses import (
+    _FIELD,
+    _FIELD_INITVAR,
+    _FIELDS,
+    _HAS_DEFAULT_FACTORY,
+    _POST_INIT_NAME,
+    MISSING,
+    _create_fn,
+    _field_init,
+    _init_param,
+    _set_new_attribute,
+    dataclass,
+    field,
+    fields,
+)
 from inspect import isclass
 from typing import Any, Generator, List, Optional, TextIO, Tuple, Union

 import numpy as np
 import yaml
```

aglavic:
I'm gonna merge the changes, if we want to modify the test to be complete we can still to it at a later stage.

Issue #67, opened 2021-12-06
Inconsistency in Orso.empty() - first column must be Qz


bmaranville:
Currently Orso.empty() creates a header object with empty.columns == [Column.empty()] == [Column(name=None)]

This is inconsistent with our agreed-upon specification, which requires that any existing columns up to length 4 must be named in this order: ["Qz", "R", "sR", "sQz"], and any columns after that can have any name.

Mitigations we can use include:

1. modify Column.empty() (or Orso.empty()?) to output columns == None instead of [Column(name=None)]
2. modify Column.empty() to output something like [Column(name="Qz")] or [Column(name="Qz"), Column(name="R")]
3. drop the rule about column order from the specification
4. not enforce the rule about column order in the schema

What do people think we should do?

andyfaff:
I'd like 3 and 4 to be non-starters. In real life would an empty file (or one without data having been filled in) ever be useful to save and recover?
If the answer is yes, and since I'd like to keep our existing specification, and things have to make sense, then would it be best to go with 1?

bmaranville:
I'm ok with 1 - it's the solution I used in the original pull request.  But I don't know how Orso.empty() is intended to be used - if users are going to get one of these and fill it out, then putting in the minimal column set (Qz, R) might be helpful.
If it's only ever going to be used by other programs as a starting point, then it doesn't really matter and we can just pick one and move on.

PR #68, opened 2021-12-06
Schema generator

bmaranville:
Generates schema, both JSON and YAML, from the class definitions in orsopy.

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -12,8 +12,6 @@ import warnings
 from collections.abc import Mapping
 from contextlib import contextmanager
 from copy import deepcopy
-from dataclasses import (_FIELD, _FIELD_INITVAR, _FIELDS, _HAS_DEFAULT_FACTORY, _POST_INIT_NAME, MISSING, _create_fn,
-                         _field_init, _init_param, _set_new_attribute, field, fields)
 from inspect import isclass
 from typing import Any, Generator, List, Optional, TextIO, Tuple, Union
 
@@ -27,6 +25,8 @@ try:
 except ImportError:
     from .typing_backport import Literal, get_args, get_origin
 
+from ..dataclasses import (dataclass, _FIELD, _FIELD_INITVAR, _FIELDS, _HAS_DEFAULT_FACTORY, _POST_INIT_NAME, MISSING, _create_fn,
+                         _field_init, _init_param, _set_new_attribute, field, fields)
 
 def _noop(self, *args, **kw):
     pass
@@ -91,36 +91,6 @@ def _custom_init_fn(fieldsarg, frozen, has_post_init, self_name, globals):
 
 
 def orsodataclass(cls: type):
-    if os.environ.get("GENERATE_SCHEMA", False) == "True":
-        from pydantic.dataclasses import dataclass as _dataclass
-        import functools
-        from typing import Dict, Any
-
-        class PydanticConfig:
-            """ for schema generation, otherwise unused """
-            @staticmethod
-            def schema_extra(schema: Dict[str, Any]) -> None:
-                for prop, value in schema.get('properties', {}).items():
-                    value.pop("title", None)
-
-                    # make the schema accept None as a value for any of the
-                    # Header class attributes.
-                    if 'enum' in value:
-                        value['enum'].append(None)
-
-                    if 'type' in value:
-                        value['anyOf'] = [{'type': value.pop('type')}]
-                        value['anyOf'].append({'type': 'null'})
-                    elif "anyOf" in value:
-                        value['anyOf'].append({'type': 'null'})
-                    # only one $ref e.g. from other model
-                    elif '$ref' in value:
-                        value['anyOf'] = [{'$ref': value.pop('$ref')}]
-
-        dataclass = functools.partial(_dataclass, config=PydanticConfig)
-    else:
-        from dataclasses import dataclass
-
     attrs = cls.__dict__
     bases = cls.__bases__
     if "__annotations__" in attrs and len([k for k in attrs["__annotations__"].keys() if not k.startswith("_")]) > 0:
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -25,8 +25,9 @@ try:
 except ImportError:
     from .typing_backport import Literal, get_args, get_origin
 
-from ..dataclasses import (dataclass, _FIELD, _FIELD_INITVAR, _FIELDS, _HAS_DEFAULT_FACTORY, _POST_INIT_NAME, MISSING, _create_fn,
-                         _field_init, _init_param, _set_new_attribute, field, fields)
+from ..dataclasses import (dataclass, _FIELD, _FIELD_INITVAR, _FIELDS, _HAS_DEFAULT_FACTORY, _POST_INIT_NAME, MISSING,
+                           _create_fn, _field_init, _init_param, _set_new_attribute, field, fields)
+
 
 def _noop(self, *args, **kw):
     pass
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -8,7 +8,7 @@ from copy import deepcopy
 from collections.abc import Mapping
 from typing import Optional, Union, List, Tuple, TextIO, Generator, Any
 from inspect import isclass
-from dataclasses import field, dataclass, fields, \
+from dataclasses import field, fields, \
     _set_new_attribute, _field_init, _create_fn, _init_param, \
     _FIELD, _FIELDS, _FIELD_INITVAR, MISSING, _HAS_DEFAULT_FACTORY, _POST_INIT_NAME
 import datetime
@@ -94,6 +94,36 @@ def _custom_init_fn(fieldsarg, frozen, has_post_init, self_name, globals):
 
 
 def orsodataclass(cls: type):
+    if os.environ.get("GENERATE_SCHEMA", False) == "True":
+        from pydantic.dataclasses import dataclass as _dataclass
+        import functools
+        from typing import Dict, Any
+
+        class PydanticConfig:
+            """ for schema generation, otherwise unused """
+            @staticmethod
+            def schema_extra(schema: Dict[str, Any]) -> None:
+                for prop, value in schema.get('properties', {}).items():
+                    value.pop("title", None)
+
+                    # make the schema accept None as a value for any of the
+                    # Header class attributes.
+                    if 'enum' in value:
+                        value['enum'].append(None)
+
+                    if 'type' in value:
+                        value['anyOf'] = [{'type': value.pop('type')}]
+                        value['anyOf'].append({'type': 'null'})
+                    elif "anyOf" in value:
+                        value['anyOf'].append({'type': 'null'})
+                    # only one $ref e.g. from other model
+                    elif '$ref' in value:
+                        value['anyOf'] = [{'$ref': value.pop('$ref')}]
+
+        dataclass = functools.partial(_dataclass, config=PydanticConfig)
+    else:
+        from dataclasses import dataclass
+
     attrs = cls.__dict__
     bases = cls.__bases__
     if '__annotations__' in attrs and \
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -25,12 +25,6 @@ try:
     from typing import Literal, get_args, get_origin
 except ImportError:
     from .typing_backport import Literal, get_args, get_origin
-# change of signature introduced in python 3.10.1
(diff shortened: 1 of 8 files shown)


bmaranville:
Tests are currently failing because of (at least?) two things: 

1. In "test_example.ort", data_source.sample.description contains a list of things that a YAML parser interprets as objects, not strings.  Are they supposed to be strings?  If so, we should wrap them in quotes.  If objects are ok, we should include that in the specification.
2. Orso.empty().columns == [Column(name=None)], which violates our column ordering specification.  See #67

andyfaff:
> In "test_example.ort", data_source.sample.description contains a list of things that a YAML parser interprets as objects, not strings. Are they supposed to be strings? If so, we should wrap them in quotes. If objects are ok, we should include that in the specification.

One of the discussions we had a couple of months ago was what to do with items contained in the ORT file that weren't in the specification. The decision there was that if there are items that aren't in the ORSO spec, and they're valid JSON, then they still need to be ingested and made available as attributes, and they shouldn't fail a validation.

If an item is in the specification, and it's not the correct type (thereby failing validation), then we have to decide if the assigned type is ok. From the draft spec:

> description:   * free text, further details of the sample, e.g. size

To me this indicates a big long string. Does the json loader parse this:

```
#         description:
#             - amb: air
#             - layer: Ni, thickness 100 nm
#             - sub: Si
``` 
into a dict of strings?

Some relevant questions:
- Why does the existing validator allow this through?
- How does the existing validator allow this through?
- Does the type annotation for description have to be developed further?

bmaranville:
I think it's making it through the current schema by matching: "type": "array", "items": {} from the "anyOf" list.  This basically means  a list with no constraints on what's in it.

the python YAML parser takes this input:
```yaml
description:
    - amb: air
    - layer: Ni, thickness 100 nm
    - sub: Si
```
and outputs 
```python
{'description': [{'amb': 'air'},
  {'layer': 'Ni, thickness 100 nm'},
  {'sub': 'Si'}]}
```

bmaranville:
I just tracked down how the unconstrained list option got added to the previous schema - it looks like it was manually added about 4 months back when we were keeping a separate set of class definitions by @andyfaff ( see d8a2297208520d0f2b0411d62a7d52f3cb88a4b4 ), where ```List[Any]``` is added to the Union of description types.

andyfaff:
I was coming up against this kind of thing quite often, as one tries to reduce inconsistency between the draft spec, the code, and the example files.
In a previous ORSO meeting we discussed sample description, which is why d8a2297 was added. People wanted to be able to do the kind of thing that was added in the `test_example.ort` in that commit (i.e. the ambient is this, the subphase is this, etc). That needed to be validated, so that's why `header_schema.py` was amended as it was.

When the specification says free text, I think it means something much more liberal than just a text string (which in validation space is not liberal at all). 

It'd be good to either fix the spec, or the code, so they agree with each other.

@jochenstahn, can you chip in here?

aglavic:
I had a look at the issue with the test_example.ort, which definitely has a wrong header.

Concerning to @andyfaff  questions.
> Some relevant questions:
> 
> * Why does the existing validator allow this through?
> * How does the existing validator allow this through?
> * Does the type annotation for description have to be developed further?

The issue here is, that the "validator" is not strictly validating but producing valid output. This is done by trying to convert the given dictionary to the specification. In the given case of "Union[str, List[str]]" it first tries to convert the input into a string, if that would fail it would try to convert to a list of strings. As everything can be converted into a string, it never falls back to list.

While I would prefer to avoid Union types as much as possible to make it easier for software to interpret the data, if we want a list of strings as option, this has to be put into the union first. We could handle strings differently, but it might be handy to automatically convert inputs like numbers into strings if the specification requires strings.

That said, the description field seems to me like an obvious free text entry, so putting string there makes sense.

Then the only question we might need to discuss together is, do we like the current behavior of the class generation. The positive of the current behavior is the ability of reading files that are not strictly according to spec but can be interpreted as such while, at the same time, generating valid file headers from user input. The downside is for users not being pointed toward possible format issues.

My proposal would be, to keep the class generation as it is with some error tolerance and add a user API functionality to run schema checks on their files/dictionaries.

bmaranville:
For the type of data_source.sample.description, allowing a list of anything seems like an arbitrary restriction (why does it have to be a list?)

If we are going to make that an unrestricted (including structure) attribute then typing.Any seems like a better choice than Union[str, List[str], List[Any], null]

arm61:
Looking at the Slack discussion from early October that was adjacent to this topic, it seems to me that `data_source.sample.description` should probably be `Union[str, List[str], List[Any], None]`. Additionally, it should maybe be changed to be called `composition` (see the orsopy channel).

bmaranville:
I see - the restriction to lists makes a lot more sense in the context of a "composition" (layers) field

aglavic:
Then the class should use:
`Optional[Union[list, str]`

As List[Any] includes List[str] so it is not changing the specification with the additional Union. As described above, the list should come first so conversion to list is tried first before falling back to str. I don't think the Any type has been implemented in _resolve_type but a simple list type should work and have the same effect.

bmaranville:
Sorry @aglavic I saw your comment after I pushed, and I will use your suggestion instead of adding List[Any]

bmaranville:
Now the only remaining failing test is because the output of Orso.empty() doesn't match our specification (it doesn't have the minimum two columns Qz and R)

Once we figure out what we want to do about that I think this can be merged.

aglavic:
I'd see three good options to deal with the columns list:

1. Make columns an optional argument with a default_factory to generate the correct columns. 
2. Make a custom empty method for the Orso class. This is most straight forward.
3. Define a new ColumnsList class that has a specific definition of empty or default, not sure how this would play with a schema. On the other hand, the class could be used for checks of the user input columns. (That said, I don't think this is a good solution.)

For 1. we'll need to update Orso.__init__ and Column:
```python
class Orso(Header):
...
    columns: Optional[List[Column]] = field(default_factory=Column._default_factory)
...
    def __init__(
...
        columns: List[Column] = None,
...
        if columns is None:
            columns=Orso.__dataclass_fields__['columns'].default_factory()
...


class Column(Header):
...
    @classmethod
    def _default_factory(cls):
        return [Column('Qz'), Column('R')]

```

For 2. we just need a change in Orso to add the custom method:
```python
    @classmethod
    def empty(cls) -> "Orso":
        """
        Create an empty instance of the ORSO header with
        all non-optional attributes as :code:`None`.

        :return: Empty Orso class.
        """
        res=super(Orso, cls).empty()
        res.columns=[Column('Qz'), Column('R')]
        return res
```

I'd vote for number 2. as it is straight forward and does not modify the class attributes.

bmaranville:
There's also the option to not add any columns by default, which would also pass validation

```python
    @classmethod
    def empty(cls) -> "Orso":
        """
        Create an empty instance of the ORSO header with
        all non-optional attributes as :code:`None`.

        :return: Empty Orso class.
        """
        res = super(Orso, cls).empty()
        res.columns = []
        return res
```

andyfaff:
@aglavic number 2 sounds ok, as does @bmaranville. It's worth considering how we construct an `OrsoDataset` though. This object is constructed with a header (`info`) and the data. An exception is raised if this object is constructed with a disagreement between the number of columns in the header and the number of columns in the data. I'm wondering if we should add an optional `column` parameter to `OrsoDataset` construction that allows for columns to be overridden in the header at this point. Or, if the column numbers aren't equal to auto-set the first 4 column names.

bmaranville:
Since there seems to be no objection to Artur's plan then, I'll implement it (initialize the first two columns as Qz and R).  Maybe we can start another ticket for the concerns raised by @andyfaff  for overriding columns when initializing the OrsoDataset class.  

So far I haven't run into any issues by adding columns to both the data and the header at the same time, then collating and initializing as an OrsoDataset, so that the number of columns always matches the data.  If a different workflow were followed there might be issues.

bmaranville:
This is ready for review, I think.

orsopy/fileio/base.py
```diff
@@ -91,6 +91,36 @@ def _custom_init_fn(fieldsarg, frozen, has_post_init, self_name, globals):
 
 
 def orsodataclass(cls: type):
+    if os.environ.get("GENERATE_SCHEMA", False) == "True":
```
aglavic, line 94: I don't have experience working with Pydantic and how to work with it, but it seems to me that creating a schema is similar to running unit test. So I would prefer having code that does these sort of changes separated from the core library as it reduces coupling.
Do you think this setup can be done in a script before importing the first orsopy.fileio modules?

orsopy/fileio/data_source.py
```diff
@@ -61,7 +61,7 @@ class Sample(Header):
     name: str
     category: Optional[str] = None
     composition: Optional[str] = None
-    description: Optional[Union[str, List[str]]] = None
+    description: Optional[Union[str, List]] = None
```
aglavic, line 64: Should be `Optional[Union[list, str]]`

orsopy/fileio/base.py
```diff
@@ -91,6 +91,36 @@ def _custom_init_fn(fieldsarg, frozen, has_post_init, self_name, globals):
...
+                        value['anyOf'].append({'type': 'null'})
+                    elif "anyOf" in value:
+                        value['anyOf'].append({'type': 'null'})
+                    # only one $ref e.g. from other model
+                    elif '$ref' in value:
+                        value['anyOf'] = [{'$ref': value.pop('$ref')}]
+
+        dataclass = functools.partial(_dataclass, config=PydanticConfig)
+    else:
+        from dataclasses import dataclass
```
aglavic, line 122: This was already imported and is thus redundant.

tools/header_schema.py
```diff
@@ -1,236 +1,98 @@
...
-
-
-@dataclass
-class Column:
+import os
+from typing import Dict, List
+from copy import deepcopy
+
+
+os.environ["GENERATE_SCHEMA"] = "True"
```
aglavic, line 10: I think instead of using this variable you could just overwrite the `dataclasses.dataclass` method here so it gets imported in orsopy.fileio.base in the modified version.

orsopy/fileio/base.py
aglavic: See comment on header_schema.py

tools/header_schema.py
aglavic: You could do something like:
```python
import dataclasses
_dataclass = dataclasses.dataclass

...

dataclasses.dataclass = functools.partial(_dataclass, config=PydanticConfig)
```

aglavic (approved):
Implementation is good, have a look if you want to move the Pydantic code as I proposed.

tools/header_schema.py
bmaranville: The approach you've described doesn't work - but I can try to find something that does.  I agree that having the logic to generate the schemas would ideally be separated from the orsopy base classes.
  
I do feel that schema generation is not similar to unit tests though; the schemas are a language-independent representation of the standard that can be (are) used in other contexts, while the unit tests have no meaning outside of orsopy.

bmaranville:
> Implementation is good, have a look if you want to move the Pydantic code as I proposed.

Because pydantic's dataclass implementation imports and calls dataclasses.dataclass directly, we cannot monkeypatch dataclasses to get what we want.

An alternative that has minimal impact on the core orsopy library is to do an indirect import of dataclass - how would you feel about something like 
```python
# orsopy/fileio/base.py
from local_dataclasses import dataclass
...
```
and then local_dataclasses.py simply imports from dataclasses, as 
```python
# orsopy/fileio/local_dataclasses.py
from dataclasses import dataclass
```

This would then be easy to monkey-patch after the fact by setting ```local_dataclasses.dataclasses = pydantic.dataclasses``` before doing schema generation.

bmaranville:
See the [schema_indirect_import branch](/reflectivity/orsopy/tree/schema_indirect_import) for an implementation of the indirect import approach described above.

aglavic:
> See the [schema_indirect_import branch](/reflectivity/orsopy/tree/schema_indirect_import) for an implementation of the indirect import approach described above.

I think this solution is good. The own dataclasses module can also deal with the fix from #72. We could also choose later to use completely our own implementation of the dataclasses functions by copying from one version of the python library to get rid of the possible issue of breaking due to API changes.

bmaranville:
I merged in the implementation from schema_indirect_import.  I need to do some formatting I guess.

bmaranville:
Ok, @aglavic the new implementation is ready to merge I think.  @andyfaff any thoughts?

andyfaff:
I'm not fully across all the meta programming here. If you're happy with it, then I think you should merge it.

arm61:
I am happy to agree with @andyfaff on this.

Issue #71, opened 2021-12-28
Problem importing orsopy in 3.10


andyfaff:
I'm [experiencing a problem](https://github.com/refnx/refnx/runs/4649235289?check_suite_focus=true) importing orsopy on Python==3.10:
```
refnx/dataset/__init__.py:2: in <module>
    from refnx.dataset.reflectdataset import ReflectDataset, OrsoDataset
refnx/dataset/reflectdataset.py:13: in <module>
    from orsopy.fileio import load_orso
/opt/hostedtoolcache/Python/3.10.1/x64/lib/python3.10/site-packages/orsopy/fileio/__init__.py:4: in <module>
    from .base import (Header, Column, Person, ValueRange, ValueVector, Value,
/opt/hostedtoolcache/Python/3.10.1/x64/lib/python3.10/site-packages/orsopy/fileio/base.py:389: in <module>
    class Value(Header):
/opt/hostedtoolcache/Python/3.10.1/x64/lib/python3.10/site-packages/orsopy/fileio/base.py:126: in __new__
    init_fun = _custom_init_fn(flds, False, has_post_init, 'self', globals())
/opt/hostedtoolcache/Python/3.10.1/x64/lib/python3.10/site-packages/orsopy/fileio/base.py:76: in _custom_init_fn
    line = _field_init(f, frozen, locals, self_name)
E   TypeError: _field_init() missing 1 required positional argument: 'slots'
```

For some reason it's only failing on Linux, not Windows or macOS. @aglavic, you'd be better placed to answer this query.

andyfaff:
I last ran this test on ~10/11 Dec, so if there hasn't been changes to an orsopy release, then it's possible that something changed in Python 3.10 since then to start causing issues.

andyfaff:
I've just confirmed that this is an issue on a clean conda env of Python3.10 in a Ubuntu VM.

bmaranville:
It is happening because of https://github.com/python/cpython/commit/d3062f672c92855b7e9e962ad4bf1a67abd4589b and I don't know if there is a super simple workaround other than try/except a function call with and without the slots argument, as the (private) function interface is changing within the active 3.10 branch.

andyfaff:
Is our implementation fragile?

aglavic:
We are using dataclasses in a way they are not intended. This therefore requires some low-level modification that use undocumented parts of the API. The only alternative I could see would be to completely use own code for this functionality, which comes at other cost as being less tested.

[...]

PR #72, opened 2021-12-30
Fix python 3.10.1 bug

aglavic:
Proposed solution to #71 that takes care of the signature change in `dataclasses._field_init`.

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -7,6 +7,7 @@ import json
 import os.path
 import pathlib
 import re
+import sys
 import warnings
 
 from collections.abc import Mapping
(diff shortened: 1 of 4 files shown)


bmaranville (approved):
I agree that this will address the issue for now, though the private method being called (not part of the public API) could be changed again at any time.  Maybe we can put in a todo to try to re-implement using public methods?

andyfaff:
The internals of dataclasses are an unknown field to me. Things breaking when Python changes their internal implementation does indicate that our implementation may be fragile. However, I think this needs to be merged coz otherwise we can't use orsopy on 3.10.

@bmaranville, do you have an idea for a re-implementation?

andyfaff:
There was a reason why we needed to reimplement those private methods, can anyone remember what it was?

andyfaff:
@aglavic, could you do another alpha release of orsopy so the version on PyPI works on 3.10?

bmaranville:
I think the group voted and it was agreed we would allow users to add extra attributes to header classes that are not in the schema (but will still be serialized and deserialized).  Dataclasses do not support this out of the box, and so we are extending/overriding the native dataclasses functions.

aglavic:
> @aglavic, could you do another alpha release of orsopy so the version on PyPI works on 3.10?

I've created PR #75 for the release, as we discussed in the last meeting. As we are still in Alpha I directly merged that PR and created the version tag to release on PyPI. You should be able to update.

PR #75, opened 2022-01-19
Prepare for release 0.0.4

aglavic:
As discussed we want to use a pull request for any orsopy release.

The main purpose is to allow review of version history/documentation updates and check if an update actually makes sense for the changes merged before.

As we are still in first alpha I will merge this directly, so this is just an example how to do it.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,28 @@
 History
 =======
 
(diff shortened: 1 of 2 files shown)


andyfaff:
Great, thanks Artur.

PR #77, opened 2022-01-27
MAINT: raise appropriate Exception if not an ORSO file

andyfaff:


bmaranville (approved):

arm61 (approved):

andyfaff:
NotOrsoFileError is less of a mouthful?

arm61:
The question is does ORSO have a file or a specification? I think we have the latter, so `Compatible` becomes necessary.

PR #78, opened 2022-02-01
Fix order of str in Union to create objects properly

aglavic:
The current version leads to wrong generation of lists as strings, this should fix it.

```python
from orsopy import fileio
fileio.Sample(name='bier', category='beverage',  description=['abc', 23.4, 2],  environment=['glass', 'table'])

output before:
>> Sample(name='bier', category='beverage', description="['abc', 23.4, 2]", environment="['glass', 'table']")

ouput after:
>> Sample(name='bier', category='beverage', description=['abc', 23.4, 2], environment=['glass', 'table'])

```

Some autoformatting was introduced by black/isort.


bmaranville:
Can you add a test that verifies the changed output indicated in your description of the pull request?

aglavic:
> Can you add a test that verifies the changed output indicated in your description of the pull request?

Yea, that's a good point. I added a test for Sample and Person that would fail the previous version.

PR #80, opened 2022-02-24
Implement YAML export with more flexibility in output layout per Header class

aglavic:
This PR implements an improved approach to perform the export to YAML. It uses a custom yaml.SafeDumper derived class to call the yaml_representer method for each Header derived item. 

This allows for a simple and transparent way to customize the way the YAML representation is layed out. The current use case is to toggle between compact and expanded representation of the dictionary to allow the Value, ValueRange and Column items to be displayed in a single line for a more compact and readable header. 

More specific classes might use it in the future for more control about e.g. sorting of items, indentation or multiline strings. It might come handy for some of the simple model description, too.

Previous empty header:
``` YAML
data_source:
  owner:
    name: null
    affiliation: null
  experiment:
    title: null
    instrument: null
    start_date: null
    probe: null
  sample:
    name: null
  measurement:
    instrument_settings:
      incident_angle:
        magnitude: null
      wavelength:
        magnitude: null
      polarization: p
    data_files: []
reduction:
  software:
    name: null
data_set: up polarization
columns:
- name: Qz
  unit: 1/angstrom
- name: R
- name: sR
```

New empty header:
``` YAML
data_source:
  owner:
    name: null
    affiliation: null
  experiment:
    title: null
    instrument: null
    start_date: null
    probe: null
  sample:
    name: null
  measurement:
    instrument_settings:
      incident_angle: {magnitude: null}
      wavelength: {magnitude: null}
      polarization: p
    data_files: []
reduction:
  software:
    name: null
data_set: up polarization
columns:
- {name: Qz, unit: 1/angstrom}
- {name: R}
- {name: sR}
```

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -331,7 +331,23 @@ class Header:
 
         :return: Yaml string
         """
-        return yaml.dump(self.to_dict(), sort_keys=False)
+        return yaml.dump(self, Dumper=OrsoDumper, sort_keys=False)
+
+    def _to_object_dict(self):
+        output = {}
+        for i, value in self.__dict__.items():
+            if i.startswith("_") or (value is None and i in self._orso_optionals):
+                continue
+            output[i] = value
+        return output
+
+    def yaml_representer(self, dumper: yaml.Dumper):
+        output = self._to_object_dict()
+        return dumper.represent_mapping(dumper.DEFAULT_MAPPING_TAG, output, flow_style=False)
+
+    def yaml_representer_compact(self, dumper: yaml.Dumper):
+        output = self._to_object_dict()
+        return dumper.represent_mapping(dumper.DEFAULT_MAPPING_TAG, output, flow_style=True)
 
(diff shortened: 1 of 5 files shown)


bmaranville:
This change looks fine to me.  I wouldn't want to go much farther to try to make the header compact (we're adding maintenance technical debt) - but this seems like a reasonable compromise and isn't a lot of extra code.

arm61:
Doesn't this make it possible to create something that doesn't *strictly* follow the specification of the ORSO file format? While it is all valid yaml it isn't the exact same as that in the specification is that something we want to enable?

bmaranville:
I don't think we can restrict the specification to a narrow subset of valid YAML representations without a tremendous amount of work...  I think we're much better off specifying the content, i.e.  "any valid YAML/JSON representation of _this information_", but making it easy in the implemented library to output a pleasing-looking version, as is done in this PR.

arm61:
That's fair and I agree. The PR looks fine to me.

PR #81, opened 2022-03-08
To nexus

bmaranville:
Adding some support for writing a nexus file (experimental).
This PR is not meant to be accepted without significant discussion and editing.

Notes:

- In nexus, information is stored in attributes or fields.  In the HDF5 file, you can store the same types of data in both places, the difference is one of convention.  In the ORSO header classes, there is no obvious hierarchy to the class attributes, so to first order they should either all be attributes or all fields in the HDF version (I chose fields)
- The data loaded by load_orso has shape (rows, columns), which makes it hard to write column-oriented data to HDF (have to loop over non-first index).  It might make more sense to have the default order of OrsoDataset.data be (columns, rows) instead. In many cases OrsoDataset.data might not be a single ndarray at all, but rather an iterable (list, tuple...) of arrays with arbitrary dimension, but the first dimension within each column should always be rows, e.g. OrsoDataset.data = [array(rows, 2, 2), array(rows), array(rows, 5)].  This would then work transparently if an ndarray of shape (columns, rows) was passed in, as numpy will happily iterate over the first dimension.
- There is no currently no linking between Orso.columns in the "header" and the actual data in the ```data``` group.  I'm not sure you need a "columns" section in the metadata of the HDF5 file if you can add all the attributes from the Orso.columns list to the data directly.  Otherwise we could just make a soft link between the Column metadata and the data in ```data```.

I'm attaching the test_example2.ort file loaded and then saved as nexus
```python
from orsopy.fileio import load_orso
from orsopy.fileio.orso import save_nexus

ex = load_orso("./test_example2.ort")
for entry in ex:
    entry.data = entry.data.T

save_nexus(ex, "test_nexus.orb")
```
[test_nexus.zip](https://github.com/reflectivity/orsopy/files/8208934/test_nexus.zip)

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -450,14 +450,6 @@ class Value(Header):
 
     yaml_representer = Header.yaml_representer_compact
 
-    def to_nexus(self, root, name):
-        group = root.create_group(name)
-        group.create_dataset("magnitude", dtype=float, data=self.magnitude)
-        if self.unit is not None:
-            print("unit", self.unit)
-            group.create_dataset("unit", data=self.unit)
-        return group
-
 
 @orsodataclass
 class ValueRange(Header):
@@ -471,14 +463,6 @@ class ValueRange(Header):
 
     yaml_representer = Header.yaml_representer_compact
 
-    def to_nexus(self, root, name):
-        group = root.create_group(name)
-        group.create_dataset("min", dtype=float, data=self.min)
-        group.create_dataset("max", dtype=float, data=self.max)
-        if self.unit is not None:
-            print("unit", self.unit)
-            group.create_dataset("unit", data=self.unit)
-        return group
 
 @orsodataclass
 class ValueVector(Header):
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -34,6 +34,8 @@ def _noop(self, *args, **kw):
     pass
 
 
+JSON_MIMETYPE = "application/json"
+
 yaml.emitter.Emitter.process_tag = _noop
 
 # make sure that datetime strings get loaded as str not datetime instances
@@ -81,11 +83,11 @@ def _custom_init_fn(fieldsarg, frozen, has_post_init, self_name, globals):
     )
 
 
-ORSO_DATACLASSES = set()
+ORSO_DATACLASSES = dict()
 
 
 def orsodataclass(cls: type):
-    ORSO_DATACLASSES.add(cls)
+    ORSO_DATACLASSES[cls.__name__] = cls
     attrs = cls.__dict__
     bases = cls.__bases__
     if "__annotations__" in attrs and len([k for k in attrs["__annotations__"].keys() if not k.startswith("_")]) > 0:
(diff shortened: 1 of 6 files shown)


bmaranville:
If we wanted to be closer to the structure in the .ort files and follow the class hierarchy, I suppose there should really be just two groups in each entry, as below.  The current save_nexus puts the ```data``` group inside the Orso (Header) group, at the same level as ```columns```

- up_up (NX_class = NXentry, ORSO_class = OrsoDataset)
  - info (ORSO_class = Orso)
    - columns
    - data_source
    - reduction
    - ...
  - data (NX_class = NXdata, no ORSO class defined yet)
    - Qz
    - R
    - sQz
    - sR
- down_down...

bmaranville:
Another note:
 - the mapping of lists of objects is awkward in HDF5, e.g. we have Orso.columns: list[Column] and Measurement.data_files: List[Union[File, str]] but HDF5 requires a unique name for children in a group.  For columns it is easy, since they are required to have a unique "name" field, so there is a rule in to_nexus that lists are converted to dict using 'name' as the key if it is available.  Otherwise, as with ```data_files```, the key is the string version of the index (data_files["0"], data_files["1"], ...)

andyfaff:
Could you make a gist and update it as you go along that would allow us to inspect the HDF produced?

bmaranville:
Sure... I didn't realize you could paste binary objects to a gist.  I will try to figure out how to do so.

bmaranville:
For the moment, here is the most recent hdf5 file created
[test_nexus.orb.zip](https://github.com/reflectivity/orsopy/files/8209524/test_nexus.orb.zip)

bmaranville:
I will do my best to keep the file test_nexus.orb in https://gist.github.com/bmaranville/aeb9bc709cf3bb1e17962ff4ae1032a5 up to date with the latest output from this fork, so it can be inspected.

arm61:
I ran this by one of the scipp developers (@SimonHeybrock who can chime in if I am talking garbage) who has spent a lot of time with NeXus. A few things he as pointed out. 
- The `R` (dependent data) should be a 'signal' (https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-signal-attribute) with the errors following the standard (https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-errors-field)
- The `Qz` should be an axes (https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-axes-attribute) with the errors following a similar standard (https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-variable-errors-field) 
- Then the units can be attributes of the datasets (`R` and `Qz`). 

Essentially the data structure should be something along the lines of 
```
data:NXdata
  @signal = "R"
  @axes = ["Qz"]
  @Qz_indices = [0]
  R: float[100]  --> the default dependent data
  errors: float[100]
  Qz: float[100]  --> the default independent data
  Qz_errors: float[100]
```

Lacking this means that the data can't really be opened by standard NeXus readers.

bmaranville:
I think we agreed that the first column would always be Q or equivalent, and so can be labelled as the x-axis for nxdata purposes, and that the second column would always be R or I and would be the Y axis, followed by optional uncertainty or resolution for (R|I), Q - so this should be straightforward since it's always the same.

bmaranville:
My attempt at handling the round-trip of lists in the orospy objects -> nexus -> orsopy objects is to put the attribute "list" = 1 on groups that are created from orosopy lists, and adding "sequence_index" to each child item in the group.  Then when being reloaded into orsopy, the "list" attribute triggers special handling that loads all the children and converts them into a list, sorting on the "sequence_index"
"sequence_index" is already a standard NeXus implementation for list items, such as in NXnote, but I believe there is no support for generic lists specifically in NeXus (NXnote classes are explicitly implied to represent a list).  In orsopy we have anonymous lists, like Orso.columns: List[Column] where the attribute "columns" doesn't have its own class, only the members of the list have an orsopy class.  Even trickier is Measurement.data_files, which is a list of either strings or File class objects.  

Handling of non-orsopy objects proceeds much like it does in writing to .ort files: they are serialized (in this case, to JSON) and saved as a field.  As a load hint, I added the attribute "mimetype": "application/json" to the fields that are being handled this way. On load, the JSON is parsed and a python dict is created from the contents for that field, which should allow for round-tripping of arbitrary user-added fields as long as they can be serialized - which is the same restriction applied to arbitrary user fields in the YAML header.  An edge case that "mostly works" in YAML but is trickier in JSON is python datetime objects, which YAML will serialize and deserialize.  The downside in the YAML case is that it will automatically deserialize any similarly-formatted string to a datetime object, even if it is intended to be a string (which would break round-tripping when a user enters a date as a string)

bmaranville:
The data in the test file, such as it is, is plottable by nexpy:

image

SimonHeybrock:
I think the uncertainties need to be stored with specific names (`errors` for the signal, and `XYZ_errors` for errors corresponding to a dataset of name `XYZ`). Otherwise we cannot identify that these datasets represent values and stddevs.

This also appears to be missing units (stored as string in `units` attributes) for all of the datasets.

bmaranville:
I just rebased this feature on the current main branch, and pushed to a new branch [save_nexus](/reflectivity/orsopy/tree/save_nexus)

bmaranville:
I am able to load the example `.ort` files, save as `.orb`, reload `.orb` into orsopy objects, and resave as `.ort` and it seems to round-trip successfully.

bmaranville:
It looks like python 3.6 is no longer available in the github CI - should we remove that from the backport testing?

andyfaff:
3.6 is really old, and out of support. I think we should drop it like a hot potato

bmaranville:
Done and done.  I updated the .yml workflow in the save_nexus branch to drop 3.6 and add 3.10 and 3.11, so we can cherry-pick the workflow file from that branch or just wait for the merge ;)

andyfaff:
Is this closable now that the other PR was merged?

bmaranville:
yes - closing it now.

PR #83, opened 2022-03-30
Model language implementation

aglavic:
Implement the ORSO header classes for the draft simple model language.
The classes include the logic for resolving names and layers to be used easily by other programs.

The [examples folder](https://github.com/aglavic/orsopy/tree/model_language/examples) includes YAML from the website examples, a script that plots reflectivity and SLD from a given sample model, a conversion script showcasing the export of GenX models to the sample model and an estimator for measurement uncertainties based on a sample model string.

Some usage of the examples:

- [simple_model_example_1.yml](https://github.com/aglavic/orsopy/blob/model_language/examples/simple_model_example_1.yml)
  `python resolve_and_plot_model.py simple_model_example_1.yml`
  res_example_1

- [simple_model_example_3.yml](https://github.com/aglavic/orsopy/blob/model_language/examples/simple_model_example_3.yml)
  `python resolve_and_plot_model.py simple_model_example_3.yml`
  res_example_3

- [simple_model_example_4.yml](https://github.com/aglavic/orsopy/blob/model_language/examples/simple_model_example_4.yml)
res_example_4

- `python sample2data.py "air | 5 (Co 31 | Fe 13) | Si"`
  sample2data

- `python simple_model_from_genx.py ...\genx\genx\genx\examples\X-ray_Reflectivity.hgx simple_model_export_genx.yml`
  => [simple_model_export_genx.yml](https://github.com/aglavic/orsopy/blob/model_language/examples/simple_model_export_genx.yml)


examples/simple_model_example_4.yml
```diff
@@ -0,0 +1,28 @@
...
+        sub_stacks:
+            LL:
+                sequence:
+                - material: {sld: 1.88401254e-06}
+                  thickness: 9.0
+                  roughness: 3.0
+                - material: {sld: -3.73401535e-07}
+                  thickness: 1.4
+                  roughness: 3.0
+                represents:  refnx.reflect.LipidLeaflet
```
andyfaff, line 14: This setup looks interesting. However, what happens if/when the parameterisation of a given function/class changes? Or more generally if there are changes between different versions of an analysis package?

examples/simple_model_example_1.yml
```diff
@@ -0,0 +1,4 @@
+sample:
+    model:
+        origin: guess based on preparation
+        stack: air | 10 ( Si 7 | Fe 7 ) | Si
```
andyfaff, line 4: This is the way that refnx specifies things, i.e. the structure is always defined as fronting --> layers --> backing, where the fronting is the medium that the radiation is incident through. I could be wrong but refl1d thinks about it from sample point-of-view. i.e. Silicon--> layers --> air. Did we normalise on a way to handle this?

examples/simple_model_example_4.yml
aglavic: That's a good point, maybe such a description would need to include the software version, e.g. in the model "origin" attribute.

As alternative it is always possible we define our own "LipidLeaflet" model class that defines the slab-generation. It was just my understanding, that there is not really a consensus in the soft-matter community on how these are parameterized.
So this was a proposal how to deal with that in a generic way.

examples/simple_model_example_1.yml
aglavic: According to Jochen the consensus was to use this convention. Which has the added advantage that a list with 1-layer per line has the correct order, too.

examples/simple_model_export_genx.yml
```diff
@@ -0,0 +1,55 @@
...
+    stack: Pt | Fe
+  buffer:
+    repetitions: 1
+    stack: bufPt | bufFe
+layers:
+  topPt:
+    thickness: 11.0
+    roughness: 3.0
+    material:
+      sld: {real: 4.864311704082625e-06, imag: -4.533346598101179e-07}
```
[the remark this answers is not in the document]
aglavic: While this is already possible by defining the correct default units through the "globals" attribute of the model, we will need to find consensus on what the default for non-provided "globals" should be.
Right now I used these:
```python
@orsodataclass
class ModelParameters(Header):
    roughness: Optional[Value] = Value(0.3, "nm")
    length_unit: Optional[str] = "nm"
    mass_density_unit: Optional[str] = "g/cm^3"
    number_density_unit: Optional[str] = "1/nm^3"
    sld_unit: Optional[str] = "1/angstrom^2"
    magnetic_moment_unit: Optional[str] = "muB"
```

jfkcooper:
Am I being slow, or is there a unit difference in the thicknesses between the examples 1/3 and 4? The total thicknesses don't seem to add up for me without 1/3 being in nm and 4 being in A?

aglavic:
@jfkcooper 
What you are seeing is the newly added automatic unit conversion in combination with the globals-length_unit arguments. So example 1 is the default, which is nm, and 3+4 should be in angstrom. 

But I noticed now a bug in the example 3, where the attribute is called length_units instead of length_unit, will correct that.

aglavic:
According to @jochenstahn it was decided to merge these developments. Will create another PR for the revision that will include the documentation and proper review.

PR #85, opened 2022-05-04
Implement changes discussed on Slack

aglavic:
As there were no further comments about my proposed changes of the class attributes on Slack I have implemented these as they have been proposed/discussed to prepare for release on the general assembly.

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -267,7 +267,10 @@ class Header:
                 if item in get_args(hint):
                     return item
                 else:
-                    raise ValueError(f"Has to be one of {get_args(hint)}")
+                    warnings.warn(
+                        f"Has to be one of {get_args(hint)} got {item}", RuntimeWarning,
+                    )
+                    return str(item)
         return None
 
     @classmethod
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -626,7 +626,7 @@ class ErrorColumn(Header):
             if self.distribution[0] == "gaussian":
                 return 1.0 / (2.0 * sqrt(2.0 * log(2.0)))
             elif self.distribution[0] == "triangular":
-                return 1.0 / sqrt(24.0)
+                return 1.0 / sqrt(6.0)
             elif self.distribution[0] == "uniform":
                 return 1.0 / sqrt(12.0)
             elif self.distribution[0] == "lorentzian":
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -613,9 +613,8 @@ class ErrorColumn(Header):
 
     error_of: str
     error_type: Optional[Literal["uncertainty", "resolution"]] = None
-    distribution: Optional[
-        Tuple[Literal["gaussian", "triangular", "uniform", "lorentzian"], Literal["sigma", "FWHM"]]
-    ] = None
+    value_is: Optional[Literal["sigma", "FWHM"]] = None
+    distribution: Optional[Literal["gaussian", "triangular", "uniform", "lorentzian"]] = None
 
     yaml_representer = Header.yaml_representer_compact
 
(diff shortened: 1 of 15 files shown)


orsopy/fileio/base.py
```diff
@@ -468,6 +485,42 @@ class Column(Header):
...
+    error_type: Optional[Literal["uncertainty", "resolution"]] = None
+    distribution: Optional[
+        Tuple[Literal["gaussian", "triangular", "uniform", "lorentzian"], Literal["sigma", "FWHM"]]
+    ] = None
+
+    yaml_representer = Header.yaml_representer_compact
+
+    def to_sigma(self):
+        """
+        Returns the multiplicative factor needed to convert a FWHM to sigma.
```
arm61, line 504: Is there a reference for these conversions? In particular the triangular and uniform ones?

arm61 (approved):

orsopy/fileio/base.py
aglavic: Well, I got them out of my memory and checked with Wikipedia, but I can see if I find a text book to reference here.

arm61: I hadn't ever thought about the conversations for triangular and uniform is all. Happy with your memory!

aglavic: I added some comments and reference now.

orsopy/fileio/data_source.py
```diff
@@ -32,8 +35,8 @@ class Experiment(Header):
 
     title: str
     instrument: str
-    start_date: str
-    probe: Union[Literal["neutrons", "x-rays"]]
+    start_date: datetime
+    probe: Literal["neutrons", "x-rays"]
```
bmaranville, line 39: The list of probes in Nexus is (from https://manual.nexusformat.org/classes/base_classes/NXsource.html?highlight=probe)

probe: (optional) [NX_CHAR](https://manual.nexusformat.org/nxdl-types.html#nx-char)
type of radiation probe (pick one from the enumerated list and spell exactly)
Any of these values:
- neutron
- x-ray
- muon
- electron
- ultraviolet
- visible light
- positron
- proton

orsopy/fileio/base.py
```diff
@@ -585,6 +605,62 @@ class Column(Header):
...
+@orsodataclass
+class ErrorColumn(Header):
+    """
+    Information about a data column.
+    """
+
+    error_of: str
+    error_type: Optional[Literal["uncertainty", "resolution"]] = None
+    distribution: Optional[
+        Tuple[Literal["gaussian", "triangular", "uniform", "lorentzian"], Literal["sigma", "FWHM"]]
```
bmaranville, line 617: Grouping the distribution type with the width specifier here seems unecessary.  Why not split out the "sigma" vs "FWHM" to another attribute?

orsopy/fileio/data_source.py
```diff
@@ -61,30 +64,31 @@ class Sample(Header):
...
+    beam, if a spin analyser is present.
+    """
+
+    unpolarized = "unpolarized"
+    p = "p"
+    m = "m"
+    mm = "mm"
+    mp = "mp"
+    pm = "pm"
+    pp = "pp"
```
bmaranville, line 87: There seemed to be general agreement on Slack that we could add ```["?m", "m?", "?p", "p?"]``` to represent half-polarized states (where the part that has unknown/unmeasured polarization is then explicit) - these would replace the bare ```l"p", "m"]``` that don't provide information about which half is unknown.

tests/test_example2.ort
```diff
@@ -63,8 +63,8 @@
 # columns:
 #     - {name: Qz, unit: 1/angstrom, dimension: WW transfer}
 #     - {name: R, dimension: reflectivity}
-#     - {name: sR, dimension: error-reflectivity}
-#     - {name: sQz, unit: 1/angstrom, dimension: resolution-WW transfer}
+#     - {error_of: R}
+#     - {error_of: Qz, error_type: resolution}
```
bmaranville, line 67: shouldn't an indication of FWHM vs sigma make it to this column descriptor?

orsopy/fileio/base.py
aglavic: Yes, good point. We separated it into "distribution" and "value_is" attributes.

orsopy/fileio/data_source.py
aglavic: Question marks would make it problematic to use as attribute names. Jochen and I preferred to go with "o" instead, hope that's OK.

orsopy/fileio/data_source.py
aglavic: Replaced neutrons and x-rays with the values here. We don't see the use of the other radiation types at the moment.

orsopy/fileio/data_source.py
bmaranville: why would they be used as attribute names, instead of values?

PR #86, opened 2022-05-05
Unit conversion for Values

aglavic:
This PR adds a ComplexValue class and implements unit conversions via pint library for all Value objects. This was partially introduced in the model language PR but I think it makes sense to add this functionality to the library, anyway.

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -194,7 +194,7 @@ class Header:
                 except ValueError:
                     # string wasn't ISO8601 format
                     return None
-            if issubclass(hint, Header) and hasattr(item, "keys"):
+            if issubclass(hint, Header) and isinstance(item, dict):
                 # convert to dataclass instance
                 attribs = hint.__annotations__.keys()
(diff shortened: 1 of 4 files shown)


orsopy/fileio/base.py
```diff
@@ -192,7 +194,7 @@ def _resolve_type(hint: type, item: Any) -> Any:
                 except ValueError:
                     # string wasn't ISO8601 format
                     return None
-            if issubclass(hint, Header):
+            if issubclass(hint, Header) and hasattr(item, "keys"):
```
arm61, line 197: Why not `isinstance(item, dict)`, which seems clearer to me?

arm61 (approved):

orsopy/fileio/base.py
aglavic: I was not sure if we can assume that the item is actually a dict and not a dict like object. This seemed to me a bit more general but I checked now and all relevant other classes like OrderedDict are actually subclasses of dict and this should work.

PR #88, opened 2022-06-08
Release a fix

aglavic:
The release did miss the schema files needed for validation of header


arm61:
Why are the tests failing...? Looks like the `orsopy/orsopy.py` file is necessary.

aglavic:
> Why are the tests failing...? Looks like the `orsopy/orsopy.py` file is necessary.

The test did nothing but import that empty module, so I removed the test.

arm61:
Fair enough, must have been a holdover from the cookie-cutter I used to start the package.

PR #92, opened 2022-06-20
Fix bug in Sample

aglavic:
Creating a Sample object that contains *sample_parameters* dictionary fails with ORSOResolveError. This was caused by the type resolution only working for Dict[key, value] or dict but not Dict.

Added a warning and a suitable type hint for sample_parameters.


arm61:
Sorry for the delay (was on annual leave), looks fine to me.

PR #94, opened 2022-07-28
Update ort spec from GA comments

aglavic:
I've implemented some additional optional attributes following the discussions at the general assembly and the anonymous feedback https://www.reflectometry.org/file_format/specs_discussion. As these only add new optional items I don't see an issue updating the spec to a new minor revision.

There is just one change that "breaks" the previous spec and that is the change of `Column.dimension` to `Column.physical_property`. For me this is an important correction of a mistake in the previous spec.

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -451,6 +451,42 @@ class ErrorValue(Header):
 
     yaml_representer = Header.yaml_representer_compact
 
+    @property
+    def sigma(self):
+        """
+        Return value converted to standard deviation.
+
+        The conversion factors can be found in common statistics and experimental physics text books or derived
+        manually solving the variance definition integral.
+        (e.g. Dekking, Michel (2005).
+        A modern introduction to probability and statistics : understanding why and how.
+        Springer, London, UK:)
+        Values and some references available on Wikipedia, too.
+        """
+        if self.value_is == "FWHM":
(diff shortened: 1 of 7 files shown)


arm61 (approved):

bmaranville:
The introduction of a new attribute name "physical_quantity" doesn't seem like the ideal solution to the problem faced.  I feel like the names described are just the canonical list of all the predefined values that can be represented either as header items in data_source.experiment.measurement or as columns.  The only items we require to be columns are Q and R/I and dR/dI.

We need to have a way to indicate if one of these quantities is going to be in that header place or in a column (could be as simple as adding "column": <columname> to the attribute in the header as a pointer in place of Value, or in the Value slot)

aglavic:
I see the `physical_quantity `attribute more as a fix of the wrong nomenclature of `dimension`, not as solution to the question of duplicates in header and column information. 

I agree, that it would be nice to have a generic indication that a column specifies additional details to a quantity in the header. Adding an attribute to the header attribute seems a good solution. I think having a column only makes sense for `ValueRanges`, so we could add that optional `column `attribute to the `ValueRange` class.

aglavic:
It seems there are no more comments to this PR, so I'll merge it at the end of the week if no one opposes.

arm61:
I am going to go ahead with this merge now.

PR #95, opened 2022-10-12
[DOC]: Simple example usage docs

arm61:
Add a simple and general example of how to use the orsopy fileio module to produce a valid .ort file


arm61:
Need to add something about `Column` for the code to be "realistic".

docs/fileio_usage.rst
```diff
@@ -0,0 +1,62 @@
...
+Usage
+=====
+
+The easiest way to use :py:mod:`orsopy.fileio` (the module of :py:mod:`orsopy` that includes file reading and writing) to produce **metadata-rich .ort reduced reflectometry files** involves integrating this into your data reduction workflow.
+Early in the workflow, the :py:mod:`orsopy.fileio` should be imported and an empty :code:`orsopy.fileio.orso.Orso` header object (here we also import :py:mod:`numpy` which will be used later). 
+
+.. code-block:: python 
+
+    import numpy as np
+    from orsopy.fileio.orso import Orso
```
aglavic, line 11: We should make use of the existing namespace in fileio, where we collect all necessary classes already. So I would propose `from orsopy import fileio` or `from orsopy.fileio import */Orso,... `

docs/fileio_usage.rst
```diff
@@ -0,0 +1,62 @@
...
+
+    header = fileio.orso.Orso.empty()
+
+Having created the empty header object we can start to populate the appropriate components of it. 
+It is generally a good idea to populate the components as particular steps occur in the reduction process. 
+For example, if we want to identify the probing radiation as neutrons, we include this as follows. 
+
+.. code-block:: python 
+
+    header.data_source.experiment.probe = 'neutrons'
```
aglavic, line 21: I think the best practice would not be to modify the empty Orso but to just use it as reference to see what is needed but to build the file directly from the classes. 
Direct assignment of attributes could be problematic as no check of correctness is done.

aglavic (approved):
I think this is a good first step. But we should just include a few example python scripts, that will be self explanatory and show a full valid file generation.

arm61:
My worry with "just include a few example python scripts" is that (for ESS reduction code) the interaction with `orsopy` is segmented over many different modules so it can be difficult to follow the thread.

docs/fileio_usage.rst
arm61: This is a good point. Let me revise the example.

bmaranville: When is a check for correctness done?

arm61: Screenshot 2022-10-18 at 09 12 58

I think he means this

tomarnoldess (approved):
I think this is good enough for now. Let's try it out on our test case at J-Parc. I will pass it on once full approved

arm61:
I still want to make it a better example of "best practice" before I merge. Working on it a bit tomorrow.

arm61:
Thoughts on having a `from_dict()` class method for `Orso`? This would mean that one can create an empty object, populate it and then do a `Orso.from_dict(my_object.to_dict())` to check everything. @aglavic @bmaranville

bmaranville:
There is already a deserializer functionality in fileio.orso, used by ```load_orso```.  How would you envision the new ```from_dict``` method would interact with or complement that?

aglavic:
@arm61 
> Thoughts on having a `from_dict()` class method for `Orso`? This would mean that one can create an empty object, populate it and then do a `Orso.from_dict(my_object.to_dict())` to check everything. 

Sorry for being late to the party. For me this method is not needed, as this is already possible/used. You just have to supply the dictionary as argument `Orso(**my_object.to_dict())`. If the object is already Orso, it makes more sense to me to supply a convenience method to run the check against schema. I think we already have a function for that.
If you think it is more intuitive, we could also supply a class method `from_dict` as you suggest that just runs the code above.

arm61:
**Finally** got back to this, I have added a correctness check at the end. I think this is the most intuitive approach (cause you can tab-autocomplete from the `header` object without having to know the exact syntax.

arm61:
Thoughts on merging this as is with the plan to improve in the future @aglavic @bmaranville ?

bmaranville:
Do we have a specification for the "physical quantity"? This definition is important to me, but I am not sure where we have written down the "official" list of physical quantity values, like "wavevector transfer" for q.

It would be nice if this list was available in the code somewhere, so that it could be *optionally* validated (optionally is very important.)

arm61:
From the specification (https://github.com/reflectivity/file_format/blob/master/specification.md#column-description) the physical quantity is to be a "plain name" I take that to mean something that the users sees as accurately describing what the column contains. 

I am hesitant to be overly prescriptive with respect to what can go in there.

bmaranville:
`physical quantity` has been suggested (I think by Jochen?) as a solution to the problem of unambiguously identifying values that might be columns or might be in the measurement.instrument_settings, e.g. wavelength.  In the context of that discussion it has been suggested that we formalize the names available: see https://github.com/reflectivity/file_format/blob/master/specs_discussion.md#reserve-key-words

I don't really care if we use `physical quantity` or something else, but we need some way to identify where to find e.g. wavelength or incident angle in a file, since sometimes that will be a column, and sometimes it will be buried deep in the header, and it's critical information in many cases.

arm61:
I agree with the concerns. I suggest we open a different issue for it though instead of getting into the weeds on this PR.

bmaranville:
You're right - it's not specific to this PR.  We can update the usage docs when the issue is resolved.

PR #98, opened 2023-02-20
Release version 1.1

aglavic:
Prepared the release of the next minor version which includes:
#83 Simple model language implementation ant examples
#94 Small updates to ORT specification

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -44,7 +44,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   backport:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         python-version: [3.6, 3.7]
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.8, 3.9, 3.10, 3.11]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


arm61 (approved):

bmaranville:
I'm getting an error in pytest from the ModelParameters class.  Specifically, it is unhappy with using a Value instance as the default value for the ModelParameters.roughness attribute.  Changing the default value to:
```python
@orsodataclass
class ModelParameters(Header):
    roughness: Optional[Value] = None
    length_unit: Optional[str] = "nm"
    mass_density_unit: Optional[str] = "g/cm^3"
    number_density_unit: Optional[str] = "1/nm^3"
    sld_unit: Optional[str] = "1/angstrom^2"
    magnetic_moment_unit: Optional[str] = "muB"
````
makes the error go away.

aglavic:
> I'm getting an error in pytest from the ModelParameters class. Specifically, it is unhappy with using a Value instance as the default value for the ModelParameters.roughness attribute. Changing the default value to:

What is your configuration? I don't get that error and all tests on github (3.6-3.9) run without errors.
I don't really understand the error message, as Value is a dataclass object and not a list or similar, so I didn't think it is considered mutable. 

We can't set the value to None, as the configuration defines the default value to be used in resolving layers. 

Anyway, I've replaced the value by a field with default_factory, that should fix your error.

bmaranville:
@aglavic the error is coming in python3.11 - they may have changed the strictness of testing for immutable defaults in that version.  You can make a dataclass immutable with a keyword in the decorator, I think, but I don't know if that will help here.

bmaranville:
As an aside, if the value can never be None, why is it typed Optional?

aglavic:
You are correct, Optional should be removed. I though I would need it because of `_orso_optionals`, but those were just needed to remove empty items on export, which can't exist if the default is different than None.

PR #100, opened 2023-02-22
add optional content hash to File object

bmaranville:
In addition to timestamp, it would be nice for reproducibility if we could specify a hash of the file contents for File objects, which are used in the Measurement class.

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -725,6 +725,16 @@ class ErrorColumn(Header):
 
 
 @orsodataclass
+class ContentHash(Header):
+    """
+    A hash of some content, using standard algorithms
+    """
+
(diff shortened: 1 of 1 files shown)


arm61 (approved):
Sounds like a good feature to me.

aglavic:
I think it's a good idea.

Some suggestions/comments:

1. Add a test for the new class.
2. Hashes are, by default, binary strings. How would we deal with that? Is it compatible with YAML? (Or did you plan to use the `hexdigest` representation, in which case the attribute should maybe be renamed.)
3. Add a convenience method in the `ContentHash `class to check a string against the given hash digest.
4. Add a convenience class mathod to `ContentHash ` create the object from the test string automatically.

PR #101, opened 2023-04-13
Convert numpy scalars to native python objects before sending to yaml dump

nvaytet:
This came up when trying to save values from Scipp variables to an Orso file.
In recent versions of Scipp, var.value() now returns numpy types, e.g. `np.float64` instead ofpython `float`.
We use the `item()` method to convert them to a python type before sending to yaml which cannot handle the numpy scalars.

### DIFF
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -431,6 +431,9 @@ class OrsoDumper(yaml.SafeDumper):
         elif isinstance(data, datetime.datetime):
             value = data.isoformat("T")
             return super().represent_scalar("tag:yaml.org,2002:timestamp", value)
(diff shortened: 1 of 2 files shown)


orsopy/fileio/base.py
```diff
@@ -431,6 +431,9 @@ def represent_data(self, data):
         elif isinstance(data, datetime.datetime):
             value = data.isoformat("T")
             return super().represent_scalar("tag:yaml.org,2002:timestamp", value)
+        elif (not np.shape(data)) and hasattr(data, "item"):
```
andyfaff, line 434: how about `elif (np.isscalar(data) and hasattr(data, "item"):`?

nvaytet: done.

PR #114, opened 2023-12-01
CI: run tests on cp312

andyfaff:


andyfaff:
@aglavic, can you have a look at this? The `orsopy` test suite doesn't run on cp312.

aglavic:
> @aglavic, can you have a look at this test fail? The `orsopy` test suite doesn't run on cp312. In general it seems quite fragile to use private methods of a package that can be subject to churn.

I'll have a look. I think this hack was necessary to allow having the *comment* attribute for derived Orso classes or something about the right yaml generation. I'll have to check and probably improve the comments to explain why this is necessary.
I remember that I didn't like going into the deep guts of dataclass but didn't see an alternative solution at that time.

PR #117, opened 2023-12-06
MAINT: make default polarization - unpolarized

### LINKED ISSUE #110: In `measurement`, `polarization` is nominally mandatory, but is not automatically included as a field
All other mandatory fields for the standard are auto populated with a `null` if left unwritten. `Polarization` is different in that it is an enum rather than a "free" field, but this is still inconsistent behaviour. not sure what the best default even could be - the obvious choice is `unpolarized` but this could be very misleading. Discussion on this is very welcome

andyfaff:
Would close #110 - if we decided it needed "fixing".

@jfkcooper 
@jochenstahn 
@aglavic 
@bmaranville

### DIFF
--- a/orsopy/fileio/data_source.py
+++ b/orsopy/fileio/data_source.py
@@ -136,7 +136,7 @@ class InstrumentSettings(Header):
     incident_angle: Union[Value, ValueRange]
     wavelength: Union[Value, ValueRange]
     polarization: Optional[Union[Polarization, ValueVector]] = field(
-        default=None,
+        default="unpolarized",
         metadata={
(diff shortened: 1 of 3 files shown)


bmaranville:
Yes, we should definitely change this default to `unpolarized`... `undefined` makes no sense at all.  If you don't know the polarization state of your beam it's like not knowing your wavelength - you can't interpret the data at all.

aglavic:
@andyfaff  I agree with the change of the default value.

However, I think this has a formal bug that needs fixing. 

`default="unpolarized",` 
will probably pass the test, but I think it should be
`default=Polarization.unpolarized`

Otherwise the value upon creation of the class will be the actual string, not the Enum, I think.

PR #123, opened 2024-02-23
DOC: try to get RTD working again

andyfaff:


andyfaff:
It might be a simple fix, in that the project config for rtd is very old.

(I have no issues if we want to host the docs elsewhere)

andyfaff:
Hmm, it looks like I don't have access to the rtd settings so it'd be difficult to investigate further. Having said that it looks like RTD is having issues connecting to github at the moment.

Issue #125, opened 2024-02-26
Specification of Column units attribute values?


bmaranville:
There is currently a test in `orsopy.fileio.tests.test_schema` that is expected to fail if the units for the first column are set to to the string: `"test_fail_unit"`

This is a problem, since in principle the Column schema needs to be applied to not just the first two columns `["Qz", "R"]`, but to any additional columns as well (after the first 4 columns, such as `wavelength`, `incident_angle`, etc.) which will not have the units specified manually in the current schema: `{"enum": ["1/angstrom", "1/nm", "1", "1/s", None]}`

We can address this in a couple ways:

1. create a separate schema for additional columns, in which `unit` is any string
2. change the Column schema so that `unit` is any string (which is the schema already defined in the Python code - the JSON schema is being manually modified to make it more restrictive in this spot)

andyfaff:
Is it possible to have the enumerated units for the first four columns, and any other string for additional columns, without creating a separate schema for those extra columns?

Will solution 2 permit any units for any column? I guess I don't mind this so much. If the default units aren't one of the enumerated ones I'd just go for the default of 1/angstrom.

Ultimately I prefer whatever solution is easiest to maintain.

aglavic:
I support solution 2 with a custom code in _read_header_data that check this additional constraint.

Lookin at it I realized that there is actually no user function to trigger the schema validation. We should consider adding it to either the Orso class directly (for check in safe and load scenarios) or to load_orso as keyword argument.

PR #126, opened 2024-02-26
Generate schema with pydantic v2

bmaranville:
Using pydantic v2.x, we can use a TypeAdapter to convert stdlib dataclasses to a pydantic model, which we can then use to generate a schema.  

This removes the need to use pydantic dataclasses, and therefore we don't need the indirect import of the dataclass decorator from a central location (which was previously monkey-patched when generating the schema)

I ran across an issue with the `Column.unit` attribute: in the old schema generator, a constraint was manually added to that attribute at schema-generation time which restricted the value to one of `["1/angstrom", "1/nm", "1", "1/s", None]`, which is ok for the `Qz` and `R` columns but which won't work for additional columns like `wavelength` (see #125)

For the moment I replicated the old behavior, with a note that it needs to be changed (existing tests all pass right now)

### DIFF
--- a/orsopy/__init__.py
+++ b/orsopy/__init__.py
@@ -1,6 +1,3 @@
 """Top-level package for orsopy."""
 
 __version__ = "1.2.0"
-
-# this is imported in submodules to be able to overwrite it with pydantic
-from dataclasses import dataclass
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -26,9 +26,7 @@ try:
 except ImportError:
     from .typing_backport import Literal, get_args, get_origin
 
-from dataclasses import MISSING, field, fields
-
-from .. import dataclass
+from dataclasses import MISSING, dataclass, field, fields
 
 
 def _noop(self, *args, **kw):
--- a/orsopy/fileio/base.py
+++ b/orsopy/fileio/base.py
@@ -2,6 +2,7 @@
 Implementation of the base classes for the ORSO header.
 """
 
+import sys
 import datetime
 import json
 import os.path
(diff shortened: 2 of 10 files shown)


andyfaff:
It seems that Pydantic V2 only supports Python3.7 and above, https://pypi.org/project/pydantic/2.0/#files.

Possible workarounds include:

- disable the test for Python 3.6.
- bump the minimum Python version to 3.7. (note that the final security fix for Python3.6 was 3 years ago).

I have a preference for the latter option but could live with the former. At some point we're going to have to bump Python versions.

andyfaff:
Otherwise I think this PR is mergeable.

I wonder if we can actually remove the generated yaml and json schemas from the repository? Having them present will lead to churn as they seem to change frequently.  Instead we could add a build step in which the schemas are generated. When wheels are generated the schemas will be present in the package.

aglavic (approved):
For me the change is fine.

Concerning the column units. I think there is no actual solution in a schema that would allow some units to be checked and others not. I personally would prefer to keep the schema the same as the class hint and to add an additional check to the validation code we use. 
This will make the schema code and result more clean and it is a special case in anyway, does nore really matter if we make it special during schema generation or during schema validation.

Concerning the schema files: I don't think we can remove them right now as the schema header points to the location to download them. It would make sense to do it in a release step but then this needs to be done in the right way that it keeps the file in the repo.
But I don't actually think this is necessary. The schema file only changes in case the schema changes AND the header_schema.py script is run. So we can just update them on the commit that is intendet for the release, which should be done in any case.

Py3.6 tests failing: I'd just remove the schema test for that version. We don't officially support it but the backport is valuable and technically still works.

andyfaff:
> Concerning the column units. I think there is no actual solution in a schema that would allow some units to be checked and others not. I personally would prefer to keep the schema the same as the class hint and to add an additional check to the validation code we use.

So allow any string for units in the schema, and add extra code to `_validate_header_data` to check the first four columns?

andyfaff:
Last commit alters the validation function to emit a `ORSOSchemaWarning` for cp36.

andyfaff:
All tests pass now (except the coverage aspect), so merging. Thanks everyone.

PR #127, opened 2024-02-28
MAINT: added separate column validation

andyfaff:
Added a separate validation for column properties. Eventually this validation should simplify the schema creation.

Instead of raising a `jsonschema.exceptions.ValidationError` I decided to raise a `ValueError` - the validation of the column properties is not being done by `jsonschema`.

Should there be extra checks?

### DIFF
--- a/orsopy/fileio/tests/test_schema.py
+++ b/orsopy/fileio/tests/test_schema.py
@@ -1,6 +1,5 @@
 import json
 import unittest
-import sys
 
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


aglavic:
Do we check anywhere, that the shape of column data is equal? If not, is this mendatory?

bmaranville:
In the `__post_init__` for `OrsoDataset`, we're checking that the number of columns in the data matches the number of columns defined in the header: https://github.com/reflectivity/orsopy/blob/main/orsopy/fileio/orso.py#L156
But we're not checking the first dimension of the columns, which should match.

What about this type of check?
```python
    def __post_init__(self):
        if len(self.data) != len(self.info.columns):
            raise ValueError("Data has to have the same number of columns as header")
        column_lengths = set(len(c) for c in self.data)
        if len(column_lengths) > 1:
            raise ValueError("Columns must all have the same length in first dimension")
```

PR #129, opened 2024-04-12
Update pytest.yml

arm61:
Remove the coveralls part of the Github action.

### DIFF
--- a/.github/workflows/docs_build.yml
+++ b/.github/workflows/docs_build.yml
@@ -27,6 +27,12 @@ jobs:
         shell: bash -l {0}
         run: |
           python setup.py install
+      - name: Generate coverage report
+        shell: bash -l {0}
+        run: |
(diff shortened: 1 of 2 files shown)


aglavic:
We could use coverage.py to run something similar as a github action. There is an output as html with detailed information that one might add to the docs, too.

I found this in a quick search: https://nedbatchelder.com/blog/202209/making_a_coverage_badge.html

bmaranville:
I tested the coverage.py solution you suggested here: https://bmaranville.github.io/orsopy/coverage/ 
It only takes a little bit of extra code in the action that builds the docs (builds the coverage document first), and we could of course add a link to the coverage report on the front page of the documentation.

I'm worried that the badge solution in that link is a little flimsy (storing state in a gist, requiring secrets, etc) and might require a lot more maintenance than the HTML output directly by coverage.py.

PR #130, opened 2024-07-01
NeXus creation fails when dictrionary of ORSO objects is present

aglavic:
Testing with some AMOR data I found a but where NeXus export failed. If there is a dictionary that contains values of Header derived type the json encoding does not work.

I've made simple fix that seems to work and produce equal result when recreating text version from it.

This PR also includes two convenience scripts that can be used in commandline:

`python -m orsopy.ort2orb *.ort`
and
`python -m orsopy.orb2ort *.orb`


bmaranville (approved):
Looks good!  The example file is very helpful.

PR #138, opened 2025-04-08
Try to reactivate python 3.6/3.7 backport tests

aglavic:
Replace runner from Ubuntu 20.04 to 22.04


aglavic:
The pyenv action does not seem to work, will look later how to fix this.

PR #148, opened 2026-01-27
Proposal for higher dimensional data and user defined spacers

aglavic:
@jochenstahn asked me about an option to add extra newlines after certain datasets to facility 2D plotting with Gnuplot of blocks of datasets (e.g. a time series as separate files that contain spin-up and spin-down data in the same time), this made me think about a possibile generic solution to higher dimensional data (off-specular / GISANS).

I've come up with a proposal of how we could handle such a situation:
- A dataset with more then 1 dimension per column is stored as flattened version (C-style indexing as it's numpy default)
- In such a case the original dimensions are stored as the optional "data_shape" attribute of the header.
- When written to .ort the last dimension is taken as "row" and every row a comment string is written with optional spacer.
- This keeps backward compatibility when reading such files. 
- 2D data can, out of the box, be plotted with Gnuplot.

In addition to this treatment, I've added some convenience functions to the OrsoDataset class that allow to iterate over columns and index it for data, in which cases the data is automatically reconstructed to the original shape. I haven't tested it yet, but I think the NeXus version should work automatically.

A second part is an optional spacer that can be added when writing text files. The class just inserts a string between datasets, but could be sub-classed for more custom behavior. Typical applications would be additional new-lines or a nice separator comment making the file more readible. It is only relevant for the text file export and lost on read.

I'll add some examples and extra tests, but wanted to get some opinions first.

Example File: [test.ort](https://github.com/user-attachments/files/24887732/test.ort.txt)

### DIFF
--- a/orsopy/fileio/__init__.py
+++ b/orsopy/fileio/__init__.py
@@ -5,7 +5,7 @@ Implementation of the Orso class that defined the header.
 from .base import (Column, ComplexValue, ErrorColumn, File, Header, Person, Value, ValueRange, ValueVector,
                    _read_header_data, _validate_header_data)
 from .data_source import DataSource, Experiment, InstrumentSettings, Measurement, Polarization, Sample
-from .orso import ORSO_VERSION, Orso, OrsoDataset, load_nexus, load_orso, save_nexus, save_orso
+from .orso import ORSO_VERSION, Orso, OrsoDataset, OrsoSpacer, load_nexus, load_orso, save_nexus, save_orso
 from .reduction import Reduction, Software
 
 __all__ = [s for s in dir() if not s.startswith("_")]
--- a/orsopy/fileio/orso.py
+++ b/orsopy/fileio/orso.py
@@ -3,7 +3,7 @@ Implementation of the top level class for the ORSO header.
 """
 
 from dataclasses import dataclass
-from typing import BinaryIO, List, Optional, Sequence, TextIO, Union
+from typing import BinaryIO, List, Optional, Sequence, TextIO, Tuple, Union
 
 import numpy as np
 import yaml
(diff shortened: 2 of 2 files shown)


andyfaff:
There are several things to unpick here.

## time based datasets
Time sliced datasets should ideally be put into separate datasets. There's no guarantee that each time slice would have the same number of data points or identical Q-values. 
Putting them into a single dataset, with an extra `time` column doesn't give the feeling of robustness.
I would say that different polarisation channels would fall into this basket.

## multidimensional datasets
<TLDR: can additional columns after Q/R/dR/dQ be multidimensional, and be stored in different HDF datasets?>

There's a high chance multidimensional datasets such as GISANS are not rectangular. e.g. whilst a detector image has a pixel grid, the Qz/Qy values probably aren't linearly spaced. For datasets like that, if you're saving in an ORT file it might be better to add columns, i.e. the regular 4, and then the extra axes.

Qz, R, dR, dQ [, Qy].

You just have to have as many rows as you do pixels.

It might be a better path to transition to ORB, rather than ORT, at this point. I believe, @bmaranville correct me if I'm wrong, that each columns doesn't have to have 1 dimension, they can be multidimensional?  If not, then I'd like for that to be worked on.

For higher dimensions binary files are probably better than text based format. It's much easier to load/save multiple dimension arrays to HDF (npy) than it is to come up with a list of rules to how they should be stored in text.

Multidimensional arrays are what is needed for sophisticated resolution smearing kernels. Here *each Q* point has an associated probability distribution, i.e. `kernel_q, p(kernel_q)` , so two arrays that have M number of points each. The mean of that probability distribution is `Qz`, and the standard deviation is `dQz`. I've been writing those to my own HDF file, but I'd like to do that with ORB.

This means that for a dataset with *N* points each of the columns has shape:

- Qz, `(N,)`
- R, `(N,)`
- dR, `(N,)`
- dQz `(N,)`    # gaussian approximation to resolution function)
- kernel_qz `(N, M)`
- kernel_pqz `(N, M)`

I reckon it's easier to do this in ORB than ORT.

andyfaff:
It might also be nice if the ORB file could have every column being multidimensional.

PR #153, opened 2026-03-16
Model language protein/RNA/DNA

aglavic:
In addition to testing all example files of the model language this PR adds the capability to resolve protein/RNA/DNA molecules directly using the SLDdb "Bio Blender" functionality. The syntax is like for ORSO database IDs.

See sample_model_example_9.yml for an example.

Will also add a local database version to perform all test locally without API calls.


bmaranville (commented):
This all looks good - I had one question about the `test` keyword in models.

PR #154, opened 2026-03-17
Separate slddb package

bmaranville:
Remove the slddb core code and import from new `slddb` package instead.  Works in concert with https://github.com/reflectivity/slddb/pull/12

(That PR will have to be merged, and the package published, before tests pass here)


bmaranville:
It might make sense to work on consolidating the logic for slddb in one place as a separate step, before considering moving it to a new home.  Marking as draft for now, in favor of #158

PR #157, opened 2026-03-26
add numpy requirement

bmaranville:
The libraries use numpy in multiple places as a top-level import, so it should really be a dependency.  

Doing `pip install orsopy` then in python doing `import orsopy.fileio` currently raises an error because of this.


aglavic (approved):
Although I think we could make the library numpy free (using a lot of simple stuff and type annotations) I don't see this as a stringent requirement. Would be surprising if someone does something with python and data without using numpy anywhere...

PR #158, opened 2026-03-26
Unified query api

bmaranville:
In this PR

- the slddb query functions are consolidated to go through `SLD_API.search(**opts)`
- the functions to support `webquery` and `blender` are moved from the SLDDB codebase to here (`orsopy.slddb`) 
- the `SLD_API.localquery` is refactored so it takes the same `qdict` argument as `webquery`
- `SLD_API.use_webquery` defaults to False
- `SLD_API.update_db` defaults to True 
- this function then routes the request to either 
   - a `webquery` to the SLDDB server (if `api.use_webquery == True`) or
   - a `localquery`, using the local copy of the `db` file

### Default configuration 
The combination of `use_webquery=False` and `update_db=True` means that, in the default configuration, this will happen:

- the first time a query is performed, the database age will be checked
- if it is older than `max_age`, a new copy is downloaded from the SLDDB server
- after that, all searches are local and use this refreshed copy of the database.

### SLDDB-server independent configuration
If you set `SLD_API.update_db = False` before your first search (typically by doing `orsopy.slddb.api.update_db = False`), then no attempt is made to download a new copy of the database, and you use the version you currently have.

### Webquery-preferred configuration
This is the currently the default in the released `orsopy` - `orsopy.slddb.api.use_webquery = True`, and it only falls back to the local db if the `webquery` fails.  Every lookup then goes to the central SLDDB server.

### DIFF
--- a/orsopy/slddb/blender.py
+++ b/orsopy/slddb/blender.py
@@ -65,7 +65,7 @@ def collect_combination(ids, name_dict):
     elements: list[Material] = []
     loaded_ids: dict[str, Material] = {}
     for id in ids:
-        if not id in loaded_ids:
+        if id not in loaded_ids:
             try:
                 entry = db.search_material(name=ExactString(name_dict[id]))[0]
             except KeyError:
@@ -114,7 +114,7 @@ def collect_blendIDs(formula):
         ID = int(pre.split('*', 1)[1].strip())
         items.append((number, ID))
     for number, ID in items:
-        if not ID in loaded_ids:
+        if ID not in loaded_ids:
             entry = db.search_material(ID=ID)[0]
             m = db.select_material(entry)
             loaded_ids[ID] = m
--- a/orsopy/slddb/blender.py
+++ b/orsopy/slddb/blender.py
@@ -1,0 +1,136 @@
+"""
+Pure calculation functions for combining biological sequences and material blends.
+No web-framework dependencies.
+"""
+
+from .dbconfig import DB_FILE
+from .database import SLDDB
+from .material import Material
+from .element_table import get_element
+from .comparators import ExactString
+
+AMINO_ABRV = {
+    "A": "Alanine",
+    "R": "Arginine",
+    "N": "Asparagine",
+    "D": "Aspartate",
+    "B": "Aspartate",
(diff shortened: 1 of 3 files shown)


bmaranville:
The corresponding PR to remove the functions that were copied from SLDDB to here has not been made yet... in the SLDDB server code they'll be replaced by imports from here.

orsopy/slddb/blender.py
```diff
@@ -0,0 +1,136 @@
...
+        number = float(pre.split('*', 1)[0].strip('(').strip())
+        ID = int(pre.split('*', 1)[1].strip())
+        items.append((number, ID))
+    for number, ID in items:
+        if ID not in loaded_ids:
+            entry = db.search_material(ID=ID)[0]
+            m = db.select_material(entry)
+            loaded_ids[ID] = m
+        elements.append(number * loaded_ids[ID])
+    result = elements[0]
```
aglavic, line 122: Can't this be simply `return sum(elements)` ?

orsopy/slddb/blender.py
```diff
@@ -0,0 +1,136 @@
...
+    for element in elements[1:]:
+        result += element
+    return result
+
+
```
aglavic, line 127: I'd propose to do this simply by using a dictionary of collect functions and .get:
```python
COLLECT_FUNCTIONS = {
  "protein": collect_protein,
  "dna": collect_dna,
  "rna": collect_rna,
  "db": collect_blendIDs,
}
```

orsopy/slddb/webapi.py
```diff
@@ -46,13 +47,13 @@ class SLD_API:
     max_age = 1
     db: SLDDB = None
 
-    def __init__(self):
-        self.first_access = True
-        self.use_webquery = True  # only try webquery once, if error occurs switch to local database
```
aglavic, line 51: Making the default just a DB download will break the SLDdb statistical functions that are used to improve search results. (I.e. return most searched and most selected materials, first.)

aglavic (commented):
Good idea to put this all in one place. I'd not remove the default webquery, though.

PR #162, opened 2026-08-03
MAINT: ensure file is closed

andyfaff:
This is needed to ensure that a resource is definitely closed. At the moment the close doesn't happen if there's an exception raised after the yield, but before the file is closed.


aglavic:
Wouldn't it be cleaner to make this directly in a context manager and with testing for the actual class? No try/except etc required.

```python
@contextmanager
def _possibly_open_file(f: Union[TextIO, str], mode: str = "wb") -> Generator[TextIO, None, None]:
    """
       ...
    """
    if not isinstance(f, str): # or alternatively isinstance(f, TextIOBase)
        yield f
    else:
        with open(f, mode) as g:
            yield g
```

andyfaff:
that looks a lot better.

andyfaff:
I applied your suggestions Artur. I changed the default open type to 'r' to make it safer, and to reflect the `open` built-in.

I checked all usages within the library and they specify the mode anyway.

